PR #3, opened 2021-06-02
Not to delete mig instance if 'CLIENT_IN_USE' when the permutation needs it anyway

omer-dayan:
Not to delete mig instance if 'CLIENT_IN_USE' when the permutation needs to create it anyway

Example:
Current config:
mig-devices:
            "1g.5gb": 2
            "2g.10gb": 1
            "3g.20gb": 1
(A process runs on a single '1g.5gb')

Desired config:
mig-devices:
            "1g.5gb": 7

What would now happen is a failure because when the mig-parted try to delete all the instances of the current-config it gets 'CLIENT_IN_USE' because of the running process on the '1g.5gb'.

However, even if it delete it successfully it would then re-create if because of the desired-config

In this PR:
If you get CLIENT_IN_USE when trying to delete mig-instance - then check if you would anyway try to create it.
If true: dont delete this instance - and dont try to create it

Signed-off-by: omer-dayan <omerxdayan1@gmail.com>

### DIFF
--- a/pkg/mig/config/config.go
+++ b/pkg/mig/config/config.go
@@ -28,7 +28,7 @@ import (
 type Manager interface {
 	GetMigConfig(gpu int) (types.MigConfig, error)
 	SetMigConfig(gpu int, config types.MigConfig) error
-	ClearMigConfig(gpu int) error
+	ClearAndGetInstancesToCreate(gpu int, desiredConfig []types.MigProfile) ([]types.MigProfile, error)
 }
 
 type nvmlMigConfigManager struct {
@@ -144,13 +144,15 @@ func (m *nvmlMigConfigManager) SetMigConfig(gpu int, config types.MigConfig) err
 	err := iteratePermutationsUntilSuccess(config, func(mps []types.MigProfile) error {
 		clearAttempts := 0
 		maxClearAttempts := 1
+		performedClearOperationSuccessfully := false
+
 		for {
 			existingConfig, err := m.GetMigConfig(gpu)
 			if err != nil {
 				return fmt.Errorf("error getting existing MigConfig: %v", err)
 			}
 
-			if len(existingConfig.Flatten()) == 0 {
+			if performedClearOperationSuccessfully || len(existingConfig.Flatten()) == 0 {
 				break
 			}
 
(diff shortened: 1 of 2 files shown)


pkg/mig/config/config.go
```diff
@@ -211,76 +215,117 @@ func (m *nvmlMigConfigManager) SetMigConfig(gpu int, config types.MigConfig) err
...
+									return desiredConfig, fmt.Errorf("error destroying Compute instance for profile '(%v, %v)': %v", j, k, ret)
+								}
+								index := getMigProfileInConfigByPosition(desiredConfig, device, gpuInfo.ProfileId, instancesToNotCreate)
+								if index != -1 {
+									instancesToNotCreate[index] = true
+									destroyGi = false
+								}
+							} else {
+								return desiredConfig, fmt.Errorf("error destroying Compute instance for profile '(%v, %v)': %v", j, k, ret)
+							}
```
klueska, line 288: This will not work in general, as the physical location of the MIG device plays a role here.

It works in your case of going from `all-balanced` to `all-1g.5gb` with a `1g.5gb` device running because no matter what, all of the new devices were going to be `1g.5gb` anyway.

If however, you were to go the other way around, the `1g.5gb` device that is "in-use" may not be the one that is allocatable in the `all-balanced` configuration.

omer-dayan: You are totally right. 

In case which the device in used must be removed in order to create other devices it will throw error in the creation stage, and all the configuration process will fail.

But in the case where the device in use is able to stay running and all the other devices are able to be created successfully its cancelling the necessity to stop the running process.

In general, it is not always that when you have a running process on the device it means you have to stop the configuration operation - there are use cases which you can still configure according to the desired config.

I think it's better than stoping the entire configuration process. 
What do you think?

klueska: Can you explain the use case here a bit better? The current model assumes you *only* reconfigure when a GPU is fully drained and has nothing running on it. I’m not comfortable tweaking that assumption unless I understand why/how you plan on using this.

elezar:
Sorry, this was closed due to changing the default branch from `master` to `main`. Also note that this a read-only mirror of https://gitlab.com/nvidia/cloud-native/mig-parted so please consider creating the PR there once the discussion above has been resolved.

PR #4, opened 2021-06-02
Keep original kubernetes components labels when reconfiguring

omer-dayan:
In the kubernetes Daemonset:

if the original kubernetes labels on the node are:
    nvidia.com/gpu.deploy.dcgm-exporter: "true"
    nvidia.com/gpu.deploy.device-plugin: "false"
    nvidia.com/gpu.deploy.gpu-feature-discovery: "true"

after the reconfiguration they are all changed to 'true' although 'nvidia.com/gpu.deploy.device-plugin' is set to false

In this PR: 
Keep the labels original value, and assign them after reconfiguration

Signed-off-by: omer-dayan <omerxdayan1@gmail.com>

### DIFF
--- a/deployments/gpu-operator/reconfigure-mig.sh
+++ b/deployments/gpu-operator/reconfigure-mig.sh
@@ -6,6 +6,10 @@ NODE_NAME=""
 MIG_CONFIG_FILE=""
 SELECTED_MIG_CONFIG=""
 
+NDP_ORIGINAL_STATE="true"
+GFD_ORIGINAL_STATE="true"
+DCGM_ORIGINAL_STATE="true"
+
 function usage() {
   echo "USAGE:"
   echo "    ${0} -h "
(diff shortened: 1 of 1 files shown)


kpouget:
thanks for the PR @omer-dayan, the patch makes sense to me! (I'm not maintener though, just giving my feedback :))

I'm just wondering about the exit when the label cannot be fetched, I don't know if the reconfiguration should be aborted because of that, or not
```
        echo "Unable to get the NVIDIA_DEVICE_PLUGIN state"
 	__set_state_and_exit "failed" 1
```

klueska:
This seems reasonable in general. I will take a closer look at it soon. @kpouget if you want to work with him to clean things up in the meantime, that would be great.

klueska:
Hello @omer-dayan can you re-propose this as an MR at:
https://gitlab.com/nvidia/cloud-native/mig-parted/

We take issues on github, but we do all development on gitlab, not github (this is just a mirror)

klueska:
I created this to supercede this PR:
https://gitlab.com/nvidia/cloud-native/mig-parted/-/merge_requests/47

I have updated it a bit, but kept @omer-dayan as the original author.

PR #25, opened 2023-08-14
fix install script

colobas:
Fix install script based on replies to issue #23

PR #28, opened 2023-10-31
Fix docker install instructions

cfhammill:
The docker instructions as written do not work, but using the pure go example below I noticed it was just missing a path element. This seems to fix it for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,12 +151,13 @@ To build from source, please follow one of the methods below.
 
 #### Use `docker` with `go install`:
 ```
-${DOCKER} run \
+docker run \
+    --rm \
     -v $(pwd):/dest \
-    golang:1.16.4 \
+    golang:1.20.1 \
     sh -c "
-    go install github.com/NVIDIA/mig-parted/cmd@latest
-    mv /go/bin/cmd /dest/nvidia-mig-parted
+    go install github.com/NVIDIA/mig-parted/cmd/nvidia-mig-parted@latest
+    mv /go/bin/nvidia-mig-parted /dest/nvidia-mig-parted
     "
 ```
 


README.md
```diff
@@ -155,7 +155,7 @@ ${DOCKER} run \
     -v $(pwd):/dest \
     golang:1.16.4 \
     sh -c "
-    go install github.com/NVIDIA/mig-parted/cmd@latest
+    go install github.com/NVIDIA/mig-parted/cmd/nvidia-mig-parted@latest
     mv /go/bin/cmd /dest/nvidia-mig-parted
```
elezar, line 159: Could it be that the `cmd` here is not correct? Running `go install` will mostlikely create a `nvidia-mig-parted` exectuable in the go path, so the following should be `mv /go/bin/nvidia-mig-parted /dest/nvidia-mig-parted`.

elezar:
Note that this is a read-only mirror and I have pushed this branch to our GitLab repo and created a Merge Request:

https://gitlab.com/nvidia/cloud-native/mig-parted/-/merge_requests/143

I do think this needs some changes and it would also be good if you sign-off your commit (with the `-s` flag).

elezar:
> thanks @elezar, I pushed a few more changes to the branch before I saw your response, there is another pull request (#25) that fixes essentially the same code in the install script. How should I go about signing the commits, I could do a squash and sign that?

It's not required to sign the commit, just sign-off as required by the DCO. You should be able to just add something like:
```
Signed-off-by: Evan Lezar <elezar@nvidia.com>
```
(with your name and email address) as the last line of the commit message. Note that each commit needs to be signed-off.

Since the commit in #25 isn't signed-off either, I would be ok with you making the changes. Alternatively we can reach out to the author to get them to do the same.

cfhammill:
Thanks @elezar, probably best to reach out to that author of #25 on the PR thread.

I've force pushed a signed-off squash commit for this PR, should be ready to go. Let me know if you'd like additional changes and if you'd prefer those squashed onto this commit.

PR #40, opened 2024-02-28
Bump project version to v0.6.0

cdesiniotis:


deployments/systemd/packages/rpm/SPECS/nvidia-mig-manager.spec
```diff
@@ -133,6 +133,14 @@ maybe_remove_hooks_symlink
 maybe_remove_config_symlink
 
 %changelog
+*  Wed Feb 28 2024 NVIDIA CORPORATION <cudatools@nvidia.com> 0.6.0-1
+- Update to latest CUDA base image 12.3.2
+- Migrate to using github.com/NVIDIA/go-nvlib
+- Bump Golang version to 1.20.5
```
tariq1890, line 139: Can we bump golang to 1.21 as go 1.20 is close to EOL?

klueska: It may not be possible at the moment due to:
https://gitlab.com/nvidia/cloud-native/mig-parted/-/merge_requests/133#note_1467441468

klueska (commented):
Is there no change in `main.go` required anymore to bump the version there?

cdesiniotis:
> Is there no change in `main.go` required anymore to bump the version there?

No longer needed since this change: https://github.com/NVIDIA/mig-parted/commit/6d18ec1b2aa4c2efb3f26c5c7b560c1c567f14a7

klueska (approved):

deployments/systemd/packages/rpm/SPECS/nvidia-mig-manager.spec
cdesiniotis: I believe the glibc issue (which is the reason why we never updated go past 1.20.5) can be avoided by installing go in the target base image. For example: https://github.com/NVIDIA/k8s-device-plugin/blob/main/deployments/container/Dockerfile.ubi8#L19-L37

This would require changes to our Dockerfiles though. Do we want to make this change and the Go version bump to be included in the 0.6.0 release?

cdesiniotis: It appears the Dockerfile changes for mig-manager were made already in this commit: https://github.com/NVIDIA/mig-parted/commit/86bb919222538840416dd4192869b46ccbe68988. However, the Dockerfiles used to build the mig-parted deb / rpm packages were not updated with this change, meaning that updating Go may cause our packages to no longer work on systems with older glibc versions.

@elezar do we want to duplicate the Dockerfile changes here as well? https://github.com/NVIDIA/mig-parted/tree/main/deployments/systemd/packages

I will proceed with the 0.6.0 release. We can look to update Go in a follow-up.

PR #41, opened 2024-02-29
Fix make targets which generate mig-parted packages

cdesiniotis:
A previous commit, https://github.com/NVIDIA/mig-parted/commit/6d18ec1b2aa4c2efb3f26c5c7b560c1c567f14a7, added a leading 'v' to our top level VERSION variable in `versions.mk`. Since then, building the mig-parted packages fails since the `PACKAGE_VERSION` (which now has a leading 'v') does not match the version string in our changelogs: https://github.com/NVIDIA/mig-parted/blob/v0.6.0/deployments/systemd/packages/Dockerfile.ubuntu#L53. 

This PR strips the leading 'v' in the `PACKAGE_VERSION` and removes the ubuntu18.04 make targets since ubuntu18.04 tags are no longer being pushed for CUDA images starting with CUDA 12.3.2.

### DIFF
--- a/deployments/systemd/packages/Makefile
+++ b/deployments/systemd/packages/Makefile
@@ -30,7 +30,7 @@ PACKAGE_VERSION := $(VERSION:v%=%)
 
 ##### Public rules #####
 
-VALID_TARGETS = tarball ubuntu18.04 ubuntu20.04 ubi8
+VALID_TARGETS = tarball ubuntu20.04 ubi8
 
 all: $(VALID_TARGETS)
 .PHONY: all $(VALID_TARGETS)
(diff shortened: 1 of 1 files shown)


deployments/systemd/packages/Makefile
```diff
@@ -45,24 +48,10 @@ tarball:
 		-v $$(pwd)/dist/$(@):/dist \
 		$(IMAGE):$(vVERSION)-$(@)
 
-ubuntu18.04:
```
elezar, line 48: One concern I have here is that this will cause the packages we build to be incompatible with older debian-based systems. Note that these are not deployment images, but the images we use to build the binaries and package them.

This should be aligned with libnvidia-container or the nvidia-container-toolkit.

deployments/systemd/packages/Makefile
```diff
@@ -25,17 +25,20 @@ REGISTRY ?= nvcr.io/nvidia
 IMAGE := $(REGISTRY)/$(NAME)
 endif
 
+# strip 'v' from version string
+PACKAGE_VERSION := $(VERSION:v%=%)
```
elezar, line 29: Out of scope for this PR: in the libnvidia-container and nvidia-container-toolkit packages, I started generating the changelog automatically as part of the build to streamline things. Not sure if that's something we want to try here.

elezar (commented):
One note about removing the ubuntu images

deployments/systemd/packages/Makefile
cdesiniotis: @elezar thanks for working on the follow-up. Are there any pending items for this PR?

deployments/systemd/packages/Makefile
```diff
@@ -45,24 +48,10 @@ tarball:
 		-v $$(pwd)/dist/$(@):/dist \
 		$(IMAGE):$(vVERSION)-$(@)
```
tariq1890, line 49: We seem to pre-pend the `v` to the VERSION var here. Does this not conflict with the following?

```
PACKAGE_VERSION := $(VERSION:v%=%)
```

cdesiniotis: `vVERSION` is a separate variable defined in the top-level `versions.mk` file and is used when we explicitly want a version string with the leading `v`. In this case, we are using it when constructing the image tag.

deployments/systemd/packages/Makefile
elezar: I think my concerns are addressed by #42.

elezar (approved):
Let's get this in for the consistent `PACKAGE_VERSION` change and then address any follow-ups in #42.

PR #52, opened 2024-04-11
replace math/rand with the rand pkg from k8s.io/apimachinery

tariq1890:
Fixes the `gosec G404` linting issues
Removed the need to add exclude directives to the golangci-lint config file

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,5 @@
 run:
-  deadline: 10m
+  timeout: 10m
 
 linters:
   enable:
@@ -20,11 +20,3 @@ linters:
 linters-settings:
   goimports:
     local-prefixes: github.com/NVIDIA/mig-parted
-
-issues:
-  exclude-rules:
-  # We use math/rand instead of crypto/rand for unique names in config tests.
-  - path: pkg/mig/config/config_test.go
-    linters:
-    - gosec
-    text: "G404"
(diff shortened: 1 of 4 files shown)


.golangci.yml
```diff
@@ -1,5 +1,5 @@
 run:
-  deadline: 10m
+  timeout: 10m
```
elezar, line 2: This change seems reasonable.

elezar (commented):
As was the case with the changes for https://github.com/NVIDIA/k8s-device-plugin/pull/629 I don't think these are required. These are in tests and we KNOW that the use of `math/rand` will not cause issues in this case.

Although the use of `k8s.io/apimachinery/pkg/util/rand` addresses the issues, it makes it seem as if the tests actually requrie this package -- instead of using the standard library -- and increases our external dependencies.

klueska:
I don't feel strongly either way, but I do agree with the sentiment of not pulling in dependencies unless absolutely necessary. Could we move to `crypto/rand` instead?

elezar:
If I recall correctly, switching to `crypto/rand` required a number of additional code changes and since these are `_tests.go` files and we're just using `rand` to generate a random integer:
```
			successAfter := rand.Intn(uniquePermutations(mc)) + 1
```
these changes are not required.

.golangci.yml
tariq1890: I'll file another PR for it

PR #53, opened 2024-04-11
Bump golang version to 1.22.2

tariq1890:


versions.mk
```diff
@@ -17,8 +17,8 @@ VERSION ?= v0.6.0
 
 vVERSION := v$(VERSION:v%=%)
 
-CUDA_VERSION := 12.3.2
-GOLANG_VERSION := 1.20.5
+CUDA_VERSION := 12.4.1
```
elezar, line 20:
```suggestion
CUDA_VERSION := 12.4.2
```

elezar (dismissed):
Please update to the `12.4.2` version string.

Also, please separate the two changes into two commits.

versions.mk
tariq1890: Typo in my commit message . The latest version is 12.4.1

tariq1890:
@elezar  Sounds good to me!

elezar (approved):
Looks good now.

I rebased and ran mod tidy.

PR #55, opened 2024-04-12
Update to incorporate go-nvml updates to expose interface types

klueska:
Closes #57


cdesiniotis (approved):
LGTM

versions.mk
```diff
@@ -18,7 +18,7 @@ VERSION ?= v0.6.0
 vVERSION := v$(VERSION:v%=%)
 
 CUDA_VERSION := 12.3.2
-GOLANG_VERSION := 1.20.5
+GOLANG_VERSION := 1.22.1
```
cdesiniotis, line 21: Just a note -- this go version bump resolves a number of high CVEs currently flagged in mig-manager ([scan results](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/cloud-native/containers/k8s-mig-manager/security?tag=v0.6.0-ubi8&architecture=arm64)).

pkg/mig/config/config_test.go
```diff
@@ -23,31 +23,33 @@ import (
...
-	n := m.nvml.(*nvml.MockLunaServer)
+	n := m.nvml.(*dgxa100.Server)
 	r1, r2 := n.Devices[gpu].SetMigMode(nvml.DEVICE_MIG_ENABLE)
 	return r1, r2
 }
 
 func TestGetSetMigConfig(t *testing.T) {
-	nvmlLib := nvml.NewMockNVMLOnLunaServer()
+	nvmlLib := dgxa100.New()
 	manager := NewMockLunaServerMigConfigManager()
```
elezar, line 48: Nit: We instantiate an `nvmlLib` twice here since we also call `dgxa100.New()` in `NewMockLunaServerMigConfigManager`. Does it make sense to inject it or return it from the function?

pkg/mig/mode/nvml.go
```diff
@@ -47,21 +47,21 @@ func NewMockNvmlMigModeManager(nvml nvml.Interface) Manager {
 
 func (m *nvmlMigModeManager) IsMigCapable(gpu int) (bool, error) {
```
elezar, line 48: Will there be a follow up to use the equivalent functionality from `go-nvlib`?

pkg/mig/mode/nvml_test.go
```diff
@@ -52,38 +52,38 @@ func NewMockNvmlLunaServer() *nvmlMigModeManager {
 
 func (d *mockNvmlA100Device) SetMigMode(mode int) (Return, Return) {
```
elezar, line 53: At which point do we replace the usages of `mockNvmlA100Device` with those from the `mock` package? (or implement them there)?

elezar (approved):
I think the initial integration looks good. There are some cases where it may make sense to use `go-nvlib` functionality, but I think this can be done as a follow-up.

The tests also call out some places where specific mock device implementations that could be consumed by other tests may be useful.

versions.mk
elezar: This was addressed in #53

pkg/mig/config/config_test.go
klueska: Possibly, but this issue predates the changes here, and I don't want to make that change as part of this PR. This PR should be scoped to only pull in changes 1:1 with the updated library.

pkg/mig/mode/nvml.go
klueska: It's probably a good idea to audit this code to see what can be used from go-nvlib (or added to go-nvlib as necessary). This code predates go-nvlib so I'm sure there's lots of things that could be simpified now.

pkg/mig/mode/nvml_test.go
klueska: What do you mean by "at which point"? At which point in the current code? They are not replacements for the ones in the `mock` package, but rather extend from them to provide logic necessary for testing the abstractions in mig-parted specifically.

cdesiniotis:
Do we want to include this in the upcoming 0.7.0 release?

pkg/mig/mode/nvml_test.go
elezar: This was unclear from my side. The question should have been whether it makes sense to move these abstractions to a common location. And if it makes sense, does it have to be done as part of this PR. I think your answer makes it clear that these are not as common as I thought on the first read-through.

elezar:
> Do we want to include this in the upcoming 0.7.0 release?

I'll leave this to @klueska to decide. I think we said that it was not a hard requirement for the 0.7.0 release.

klueska:
I think we should explicitly leave it *out* of the imminent release (in case there are any unexpected issues it introduces).

klueska:
This is now ready from my perspective. Please review again.

elezar (approved):
lgtm.

PR #61, opened 2024-04-19
bump golang version in mig-manager dockerfiles

tariq1890:
Since we've marked the required go toolchain version as "1.22.2" in `go.mod`, I think it would be consistent to build the mig-parted binary with the same version


elezar (approved):
Although I'm not opposed to that change, I just want to point out that this value is already overridden to `1.22.2` in our make process.

PR #62, opened 2024-04-19
Bump project version to v0.7.0

cdesiniotis:


cdesiniotis:
Upon further thought, we could name this release `v0.6.1` considering no new features were introduced. The `release-0.6` branch has not diverted from `main`, so we could simply rebase `release-0.6` and then create the `v0.6.1` tag.

deployments/container/nvidia-mig-manager-example-hopper.yaml
```diff
@@ -23,7 +23,7 @@ spec:
       serviceAccountName: nvidia-mig-manager-service-account
       containers:
       - name: nvidia-mig-manager
-        image: nvcr.io/nvidia/cloud-native/k8s-mig-manager:v0.6.0-ubi8
+        image: nvcr.io/nvidia/cloud-native/k8s-mig-manager:v0.7.0-ubi8
```
tariq1890, line 26: Not a blocking comment. But we should consider publishing ubi9 images at some point

tariq1890 (approved):

deployments/container/nvidia-mig-manager-example-hopper.yaml
elezar: Do you mean that we need to replace the images? 

Note that since we copy the `mig-parted` executable from the container to the host, we want to ensure that we're using as low a base version as possible to avoid issues with GLIBC dependencies.

tariq1890: I meant use `ubi9` as the base image instead of `ubi8`

PR #63, opened 2024-04-23
Parallelize tests to allow us to reenable testing in pkg/...

klueska:


pkg/mig/config/config_test.go
```diff
@@ -68,11 +63,21 @@ func TestGetSetMigConfig(t *testing.T) {
 	}()
 
 	for _, tc := range testCases {
+		tc := tc // to allow us to run parallelly
```
ArangoGutierrez, line 66: See what we did here -> https://github.com/NVIDIA/k8s-device-plugin/blob/main/internal/plugin/server_test.go#L134

klueska: I see you storing a copy of the variable same as I do, but you don't seem to call `t.Parallelize()` anywhere. What are you trying to show me exactly?

ArangoGutierrez: I am ok with the `t.Parallel()` part, just wanted to point out the variable copy

When reviewing Tariq's PR, he wanted initially to do as you are proposing here, but after the review process we went for the second option of using the index on the array.

```go
	for i := range testCases {
		tc := testCases[i]
```

But as we commented on Tariq's PR, both approaches have the same end result.

klueska: That is a air comment and worth changing for readability. Though I'd argue in the PR you referenced, its not actually necessary to store a local copy of the variable since none of the test cases are running in parallel.

klueska:
All of the unit tests are passing locally, but one is failing on github. Investigating ...

pkg/mig/config/config_test.go
```diff
@@ -104,13 +109,16 @@ func TestClearMigConfig(t *testing.T) {
 		return testCases
 	}()
 
-	for _, tc := range testCases {
+	for i  range testCases {
```
elezar, line 112:
```suggestion
	for i  := range testCases {
```

klueska: Fixed

elezar (commented):
As a matter of interest, are there particular tests that take a long time, or why is the parallelisation requried?

klueska:
Yes, the tests were (locally) taking over 20 minutes to complete and now they finish in under 1

klueska:
OK, tests are passing now

pkg/mig/config/config_test.go
klueska: Please resolve  this comment if you are happy with the changes

ArangoGutierrez (approved):

PR #66, opened 2024-05-03
add support for H100 NVL 94GB

tariq1890:
Config added based on the docs provided [here](https://docs.nvidia.com/datacenter/tesla/mig-user-guide/index.html#h100-profiles)


ArangoGutierrez (commented):
Looks good to me, but I'll leave the final approval to @klueska

deployments/systemd/config-default.yaml
```diff
@@ -151,7 +151,55 @@ mig-configs:
...
+      devices: all
+      mig-enabled: true
+      mig-devices:
+        "1g.11gb+me": 1
+
+    - device-filter: "0x232110DE"
+      devices: all
+      mig-enabled: true
+      mig-devices:
+        "1g.22gb": 4
```
klueska, line 172: 2 out of 3 of these seem to be missing the top-level name

tariq1890: oh thank you for catching that

PR #69, opened 2024-05-10
Add dev root option to mig-manager container

cdesiniotis:


deployments/container/reconfigure-mig.sh
```diff
@@ -525,28 +541,31 @@ if [ "${?}" != "0" ]; then
...
-	echo "Running nvidia-smi"
-	chroot ${DRIVER_ROOT_CTR_PATH} nvidia-smi >/dev/null
-	if [ "${?}" != "0" ]; then
-		exit_failed
-	fi
-
 	echo "Creating NVIDIA control device nodes"
-	nvidia-ctk system create-device-nodes --control-devices --driver-root=${DRIVER_ROOT_CTR_PATH}
+	nvidia-ctk system create-device-nodes \
+	  --load-kernel-modules \
```
cdesiniotis, line 546: This won't work currently since the toolkit code to load kernel modules relies on chroot'ng into the driverRoot.

General question -- does anyone remember why it was required to run nvidia-smi first before creating the device nodes and generating the management CDI spec?

deployments/container/reconfigure-mig.sh
```diff
@@ -525,28 +541,31 @@ if [ "${?}" != "0" ]; then
...
+	  --control-devices \
+	  --driver-root=${DRIVER_ROOT_CTR_PATH} \
+	  --dev-root=${DEV_ROOT_CTR_PATH}
```
cdesiniotis, line 549: This assumes we add a `--dev-root` option to the `nvidia-ctk system create-device-nodes` command.

deployments/container/reconfigure-mig.sh
```diff
@@ -52,6 +54,8 @@ function usage() {
   echo "    -p <default-gpu-clients-namespace>            Default name of the Kubernetes Namespace in which the GPU client Pods are installed in"
   echo "    -t <driver-root>                              Root path to the NVIDIA driver installation"
   echo "    -a <driver-root-ctr-path>                     Root path to the NVIDIA driver installation mounted in the container"
+  echo "    -b <dev-root>                                 Root path to the NVIDIA device nodes"
```
cdesiniotis, line 57: Corresponding options would also need to be added to the mig-manager go binary which invokes this shell script.

deployments/container/reconfigure-mig.sh
cdesiniotis: I have reverted this change for now, and instead have resorted to setting `--driver-root=${DEV_ROOT_CTR_PATH}` since `nvidia-ctk system create-device-nodes` currently creates device nodes there.

deployments/container/reconfigure-mig.sh
cdesiniotis: I have removed `--load-kernel-modules` for the time being since this currently does not work when driverRoot is not chroot'ble.

elezar: I think we use `nvidia-smi` to create the NON control device nodes. I don't believe it's required to run `nvidia-smi` BEFORE we run this command, but we do need to run it in addition to running this if we wanted the device nodes to be created.

The issue with `nvidia-smi` is that it doesn't create the `/dev/nvidia-uvm*` devices nodes, which is why we create them here.

cdesiniotis: The non-control devices should typically exist at this point, except in the scenario where a reboot is required to toggle the MIG mode and the driver is preinstalled. Even though the driver and toolkit validations run before mig-manager starts, the invocation of nvidia-smi will create the device nodes in the container's dev and not the host's.

Any recommendations for how we should handle this case?

cdesiniotis: I have updated the script to run `nvidia-smi` first if the NVIDIA driver is installed on the host (e.g. at `/`). This should address the scenario I described in https://github.com/NVIDIA/mig-parted/pull/69#discussion_r1625372657

versions.mk
```diff
@@ -24,4 +24,4 @@ BUILDIMAGE ?=  ghcr.io/nvidia/k8s-test-infra:$(BUILDIMAGE_TAG)
 
 GIT_COMMIT ?= $(shell git describe --match="" --dirty --long --always --abbrev=40 2> /dev/null || echo "")
 
-NVIDIA_CTK_VERSION := v1.14.6
+NVIDIA_CTK_VERSION := v1.15.0
```
elezar, line 27: Do we need unreleased features for this?

deployments/container/reconfigure-mig.sh
```diff
@@ -525,27 +541,36 @@ if [ "${?}" != "0" ]; then
 fi
 
 if [ "${CDI_ENABLED}" = "true" ]; then
-	echo "Running nvidia-smi"
-	chroot ${DRIVER_ROOT_CTR_PATH} nvidia-smi >/dev/null
-	if [ "${?}" != "0" ]; then
-		exit_failed
+	if [ "${DRIVER_ROOT_CTR_PATH}" = "/host" ]; then
```
elezar, line 544: Does the `/run/nvidia/driver` case still work as expected?

deployments/container/reconfigure-mig.sh
```diff
@@ -525,27 +541,36 @@ if [ "${?}" != "0" ]; then
...
+		if [ "${?}" != "0" ]; then
+			exit_failed
+		fi
 	fi
 
 	echo "Creating NVIDIA control device nodes"
-	nvidia-ctk system create-device-nodes --control-devices --driver-root=${DRIVER_ROOT_CTR_PATH}
+	nvidia-ctk system create-device-nodes \
+		--control-devices \
+		--driver-root=${DEV_ROOT_CTR_PATH}
```
elezar, line 555: Do we want to merge https://github.com/NVIDIA/nvidia-container-toolkit/pull/526 first and switch to `--dev-root` here?

elezar (approved):
I have some minor questions in terms of the `nvidia-ctk` interaction, but these are not blockers.

deployments/container/reconfigure-mig.sh
cdesiniotis: Updated to use `--dev-root`.

versions.mk
cdesiniotis: Updated to `v1.16.0-rc.1`.

deployments/container/reconfigure-mig.sh
cdesiniotis: Base on my initial testing, yes. But will keep investigating this.

cdesiniotis: I was wrong -- the `/run/nvidia/driver` case was not working with this change. As suspected, without running `nvidia-smi` the `nvidia-cap*` device nodes were note created / updated correctly after a MIG configuration, and thus, the CDI specifications were not accurate. I have raised https://github.com/NVIDIA/mig-parted/pull/82 to address both the `/run/nvidia/driver` case as well as the GKE case, where driverRoot != devRoot.

PR #70, opened 2024-05-15
Align nvidia driver root envars

elezar:
Align the envvars used to specify the driver root with the NVIDIA Container Toolkit (NVIDIA_DRIVER_ROOT).

### DIFF
--- a/cmd/nvidia-mig-manager/main.go
+++ b/cmd/nvidia-mig-manager/main.go
@@ -47,7 +47,7 @@ const (
 	DefaultHostMigManagerStateFile   = "/etc/systemd/system/nvidia-mig-manager.service.d/override.conf"
 	DefaultHostKubeletSystemdService = "kubelet.service"
 	DefaultGPUClientsNamespace       = "default"
-	DefaultDriverRoot                = "/run/nvidia/driver"
+	DefaultNvidiaDriverRoot          = "/run/nvidia/driver"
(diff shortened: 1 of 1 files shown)


klueska (approved):

cmd/nvidia-mig-manager/main.go
```diff
@@ -211,11 +211,10 @@ func main() {
 		},
 		&cli.StringFlag{
 			Name:        "driver-root",
-			Aliases:     []string{"t"},
-			Value:       DefaultDriverRoot,
-			Usage:       "Root path to the NVIDIA driver installation. Only used if --cdi-enabled is set.",
```
klueska, line 216: Why do you drop the usage string?

cmd/nvidia-mig-manager/main.go
```diff
@@ -211,11 +211,10 @@ func main() {
 		},
 		&cli.StringFlag{
 			Name:        "driver-root",
```
klueska, line 213: If we consider the new value the "correct" value, should we change this to `nvidia-driver-root` and make the alias `driver-root`?

cmd/nvidia-mig-manager/main.go
elezar: Not intentional ... will readd.

cmd/nvidia-mig-manager/main.go
elezar: Yes. We should.

PR #85, opened 2024-06-24
[mig-manager] move library/nvidia-smi discovery logic to Go

cdesiniotis:


cmd/nvidia-mig-manager/main.go
```diff
@@ -70,6 +71,8 @@ var (
 	driverRootCtrPath string
 	devRoot           string
 	devRootCtrPath    string
+	driverLibraryPath string
+	nvidiaSMIPath     string
```
elezar, line 75: I understand that we've been using globals for the other values here, but let's rather return these from the function where we extract them.

cmd/nvidia-mig-manager/main.go
```diff
@@ -338,7 +384,7 @@ func runScript(migConfigValue string) error {
 		"-p", defaultGPUClientsNamespaceFlag,
 	}
 	if cdiEnabledFlag {
-		args = append(args, "-e", "-t", driverRoot, "-a", driverRootCtrPath, "-b", devRoot, "-j", devRootCtrPath)
+		args = append(args, "-e", "-t", driverRoot, "-a", driverRootCtrPath, "-b", devRoot, "-j", devRootCtrPath, "-l", driverLibraryPath, "-q", nvidiaSMIPath)
```
elezar, line 387: the new parameters should be arguments to this function.

cmd/nvidia-mig-manager/main.go
```diff
@@ -298,6 +306,44 @@ func start(c *cli.Context) error {
...
+// it is required to run nvidia-smi to create the nvidia-cap* device nodes.
+// If driverRoot != devRoot, we must discover the paths to libnvidia-ml.so.1 and
+// nvidia-smi in order to run nvidia-smi. We discover the paths here once and
+// pass these as arguments to reconfigure-mig.sh
+//
+// Currently, driverRoot != devRoot only when devRoot='/'. Since mig-manager
+// has rw access to the host rootFS (at hostRootMountFlag), reconfigure-mig.sh
+// will first chroot into the host rootFS before invoking nvidia-smi, so the
+// device nodes get created at '/dev' on the host.
+func setupRequiredForCDI() error {
```
elezar, line 322: This name seems "off" we're not setting anythinng up.

Could we change this to:
```suggestion
func getPathsForCDI() (string, string, error) {
```
I would even go so far as to say that we want: `getDriverPaths() (string, string, error)` that actually contains the path logic and we can call this from `getPathsForCDI`.

deployments/container/reconfigure-mig.sh
```diff
@@ -111,7 +115,13 @@ while getopts "hrden:f:c:m:i:o:g:k:p:t:a:b:j:" opt; do
...
       DEV_ROOT_CTR_PATH=${OPTARG}
       ;;
-    \? ) echo "Usage: ${0} -n <node> -f <config-file> -c <selected-config> -p <default-gpu-clients-namespace> [-e -t <driver-root> -a <driver-root-ctr-path> -b <dev-root> -j <dev-root-ctr-path>] [ -m <host-root-mount> -i <host-nvidia-dir> -o <host-mig-manager-state-file> -g <host-gpu-client-services> -k <host-kubelet-service> -r -s ]"
+    l ) # process option l
+      DRIVER_LIBRARY_PATH=${OPTARG}
+      ;;
+    q ) # process option q
+      NVIDIA_SMI_PATH=${OPTARG}
+      ;;
+    \? ) echo "Usage: ${0} -n <node> -f <config-file> -c <selected-config> -p <default-gpu-clients-namespace> [-e -t <driver-root> -a <driver-root-ctr-path> -b <dev-root> -j <dev-root-ctr-path> -l <driver-library-path> -q <nvidia-smi-path> ] [ -m <host-root-mount> -i <host-nvidia-dir> -o <host-mig-manager-state-file> -g <host-gpu-client-services> -k <host-kubelet-service> -r -s ]"
```
elezar, line 124: Any reason why we don't call `usage` and exit here? (like we do for `h`)

elezar (changes requested):
Thanks @cdesiniotis.

It would be better if we move away from global variables here -- especially for the new code.

cmd/nvidia-mig-manager/main.go
cdesiniotis: Updated.

cmd/nvidia-mig-manager/main.go
cdesiniotis: Done.

cmd/nvidia-mig-manager/main.go
cdesiniotis: Agreed. Updated to `getPathsForCDI`.

deployments/container/reconfigure-mig.sh
cdesiniotis: I am not aware of any reason not to. I can update this to call `usage` and `exit` in a follow-up.

elezar (approved):
Looks good now.

PR #90, opened 2024-07-03
[mig-manager] Add --nvidia-cdi-hook-path flag

cdesiniotis:
This provides the flexibility to support custom install directories configured with the toolkit-container, e.g. `/home/kubernetes/bin/nvidia` on GKE.

### DIFF
--- a/deployments/container/reconfigure-mig.sh
+++ b/deployments/container/reconfigure-mig.sh
@@ -32,13 +32,14 @@ DEV_ROOT=""
 DEV_ROOT_CTR_PATH=""
 DRIVER_LIBRARY_PATH=""
 NVIDIA_SMI_PATH=""
+NVIDIA_CDI_HOOK_PATH=""
 
 export SYSTEMD_LOG_LEVEL="info"
 
 function usage() {
   echo "USAGE:"
   echo "    ${0} -h "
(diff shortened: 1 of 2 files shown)


deployments/container/reconfigure-mig.sh
```diff
@@ -599,7 +608,7 @@ if [ "${CDI_ENABLED}" = "true" ]; then
 		--dev-root=${DEV_ROOT_CTR_PATH} \
 		--vendor="management.nvidia.com" \
 		--class="gpu" \
-		--nvidia-ctk-path="/usr/local/nvidia/toolkit/nvidia-ctk" | \
+		--nvidia-cdi-hook-path=${NVIDIA_CDI_HOOK_PATH} | \
```
klueska, line 611: This is a fairly recent edition. Is there a reason to use this instead of `nvidia-ctk cdi ...`?

cdesiniotis: No particular reason. I just thought to pick up the new binary since it will be used moving forward. @elezar @klueska would you prefer if I stuck with `--nvidia-ctk-path` instead?

elezar: For the record, this can be `nvidia-ctk` or `nvidia-cdi-hook` and the generated hook should be correct. 

For clarification: We want to separate the hook implementations from the `nvidia-ctk` to make them more easily auditable and / or consumable from non-OCI complaint environments such as LXD and/or Charliecloud for example.

elezar: I don't have a real preference, but in newer versions of the container toolkit we prefer `nvidia-cdi-hook`. Can we be assured that users are using a v1.16.0 toolkit?

cdesiniotis: We can't guarantee users are using v1.16.0 of the toolkit, but it will be the default version in the next operator release.

elezar (approved):

PR #96, opened 2024-07-17
Add tooling to bump golang version

elezar:


.github/workflows/golang.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
     - name: Get Golang version
       id: vars
       run: |
-        GOLANG_VERSION=$( grep "GOLANG_VERSION :=" versions.mk )
+        GOLANG_VERSION=$(./hack/golang-version.sh)
```
tariq1890, line 39: nit: I've never quite understood the convention behind naming the scripts directory `hack`. I assume that it is a practice popularised by the k8s community.

I've always preferred `scripts` as it's more intuitive to the reader

elezar: I think one could argue that `scripts` are for external consumption whereas `hack` is for internal consumption.

And yes, this seems to be a k8s convention.

PR #112, opened 2024-08-20
fix govet issue and pin golangci-lint version for CI checks

tariq1890:
This PR fixes a go vet lint issue which started to surface in the main branch on the latest golangci-lint version. Since we don't pin the version of golangci-lint in our GitHub action, there is no guarantee of CI check stability as the latest stable release of golangci-lint is always pulled at the time of running the CI. This PR solves this problem by pinning the version of golangc-lint so that the CI checks are stable and consistent

### DIFF
--- a/.github/workflows/golang.yaml
+++ b/.github/workflows/golang.yaml
@@ -45,7 +45,7 @@ jobs:
     - name: Lint
       uses: golangci/golangci-lint-action@v6
       with:
-        version: latest
+        version: v1.60.1
         args: -v --timeout 5m
         skip-cache: true
     - name: Check golang modules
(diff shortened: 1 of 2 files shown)


klueska (approved):

cdesiniotis:
@tariq1890 thanks for fixing this. I do see some benefits of using the latest version of golangci-lint in our CI -- 1) we do not need to maintain this version, and 2) we learn of new issues as golangci-lint is improved.

tariq1890:
I agree that unpinning the golangci-lint version helps us learn of new issues, however, I don't think it should cause CI failure. A CI failure should be a signal of a possibly bad change that is introduced through a changes directly introduced by the commit author. In this case, we are getting CI failures that don't pertain to any change introduced by a committer.

I think we should use a tool like `dependabot` or `renovate-bot` to automate the upgrade of golangci-lint in GitHub Actions files

Issue #120, opened 2024-09-04
Add support for the NVIDIA H200 GPU


KeithMnemonic:
The latest version (.0.9.0) does not have the configuration information for the NVIDIA H200 GPU

The `/etc/nvidia-mig-manager/config-default.yaml` file does not contain an entry for the NVIDIA H200 GPU.

```
root@viking-prod-525-10-176-19-53:~# grep 0x233510DE /etc/nvidia-mig-manager/config-default.yaml
root@viking-prod-525-10-176-19-53:~# echo $?
1
```

Here is the relevant data for the NVIDIA H200 GPU needed for  `/etc/nvidia-mig-manager/config-default.yaml` 
```
root@viking-emr-prod-277:/etc/nvidia-mig-manager#  nvidia-smi -q -i 0 | grep -i "device id" | awk '{print $4}'
0x233510DE
```

```
root@viking-emr-prod-277:/etc/nvidia-mig-manager# sudo nvidia-smi -q | grep -E "Product Name|Product Architecture|Serial Number|Board Part Number|GPU Part Number|Device Id"

    Product Name                          : NVIDIA H200
    Product Architecture                  : Hopper
    Serial Number                         : 1641824000241
    Board Part Number                     : 699-2G520-0280-RC3
    GPU Part Number                       : 2335-895-A1
        Device Id                         : 0x233510DE
```

```
root@viking-prod-525-10-176-19-53:~# nvidia-smi mig -lgip
+-----------------------------------------------------------------------------+
| GPU instance profiles:                                                      |
| GPU   Name             ID    Instances   Memory     P2P    SM    DEC   ENC  |
|                              Free/Total   GiB              CE    JPEG  OFA  |
|=============================================================================|
|   0  MIG 1g.18gb       19     7/7        16.00      No     16     1     0   |
|                                                             1     1     0   |
+-----------------------------------------------------------------------------+
|   0  MIG 1g.18gb+me    20     1/1        16.00      No     16     1     0   |
|                                                             1     1     1   |
+-----------------------------------------------------------------------------+
|   0  MIG 1g.35gb       15     4/4        32.25      No     26     1     0   |
|                                                             1     1     0   |
+-----------------------------------------------------------------------------+
|   0  MIG 2g.35gb       14     3/3        32.25      No     32     2     0   |
|                                                             2     2     0   |
+-----------------------------------------------------------------------------+
|   0  MIG 3g.71gb        9     2/2        69.50      No     60     3     0   |
[...]

PR #121, opened 2024-09-04
Add H200 with PCI ID 0x233510DE

KeithMnemonic:
This PR updates the config-default.yaml to include the GPU ID for the NVIDIA H200.
In addition, other changes were added to sync this file with the

`assets/state-mig-manager/0400_configmap.yaml`

in https://github.com/NVIDIA/gpu-operator/

These files should contain the same information but have drifted over time.

Fixes: #120

### DIFF
--- a/deployments/systemd/config-default.yaml
+++ b/deployments/systemd/config-default.yaml
@@ -4,6 +4,11 @@ mig-configs:
     - devices: all
       mig-enabled: false
 
+  all-enabled:
+    - devices: all
+      mig-enabled: true
+      mig-devices: {}
+
   # A100-40GB, A800-40GB
   all-1g.5gb:
     - devices: all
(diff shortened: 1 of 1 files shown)


deployments/systemd/config-default.yaml
```diff
@@ -210,20 +240,54 @@ mig-configs:
...
+    - devices: all
+      mig-enabled: true
+      mig-devices:
+        "4g.71gb": 1
+
+  all-7g.141gb:
+    - devices: all
+      mig-enabled: true
+      mig-devices:
+        "7g.141gb": 1
```
cdesiniotis, line 293: nit: no newline

KeithMnemonic: added

klueska:
It's hard to tell what is net-new vs. just refactoring / consolidating with the GPU operator. Can you separate this into two commits (one for consolidating with the GPU operator., and one for adding the new GPU)?

KeithMnemonic:
> It's hard to tell what is net-new vs. just refactoring / consolidating with the GPU operator. Can you separate this into two commits (one for consolidating with the GPU operator., and one for adding the new GPU)?

If the contents of the file are meant to be identical other than a few lines at the top and the tabs, is a diff of the file from the GPU operator not sufficient?   There were other missing things that were in the `configmap` that were not in the `config-default` as well that I added in.  I can send you the configmap .yaml I used

klueska:
If @cdesiniotis is confident that the diff is accurate and that we now have parity with the operator, then that is good enough for me. I'll let him do the final sign-off.

cdesiniotis:
I've downloaded the config file from this PR and the latest config file included in the gpu-operator repository and confirmed the content is the same.

cdesiniotis (approved):

PR #126, opened 2024-09-06
Update README.md

KeithMnemonic:
This PR adds a note to the README.md about a possible warning message that might appear when running
nvidia-mig-parted.


deployments/systemd/README.md
```diff
@@ -85,6 +85,19 @@ nvidia-mig-parted apply -f /etc/nvidia-mig-manager/config.yaml -k /etc/nvidia-mi
 ```
 (An alternative is to ensure that the required environment variables are passed to `sudo` by using `sudo -E` instead)
 
+**NOTE:** When running the nvidia-mig-parted, if a message appears that is similar to
```
elezar, line 88: Question: Do we want to add a `Troubleshooting` section to the readme to collect this type of information?

KeithMnemonic: I am fine with a Trobuleshooting section

PR #133, opened 2024-10-02
use the UBI9 cuda base image

tariq1890:


.gitlab-ci.yml
elezar, line 1: Out of scope for this PR, but I believe we can remove this file now that we're not using the GitLab CI.

elezar (approved):
Looks good. Thanks @tariq1890.

PR #136, opened 2024-10-15
bump default DOCKER ARG default golang version to 1.22.8

tariq1890:


cdesiniotis:
The version of Golang being used during build is actually 1.23.2: https://github.com/NVIDIA/mig-parted/actions/runs/11350079065/job/31567676999#step:7:20 which is defined here: https://github.com/NVIDIA/mig-parted/blob/main/deployments/devel/Dockerfile#L17

tariq1890:
Yes, I am waiting till 1.23.3 is released as it has the fixes for when running in a docker arm64 container

tariq1890:
I think it's good to atleast bump this default even though the actual golang version is `1.23.2`

cdesiniotis:
The commit title is misleading since you are not actually changing the version of Go used to build this project.

tariq1890:
Please check now

cdesiniotis (approved):

PR #192, opened 2025-04-08
Bump golang to 1.23.5 in the systemd service

mresvanis:
This change bumps the golang version used to build the `nvidia-mig-parted` binary for the `nvidia-mig-manager` systemd service, as it currently leads to a `go install` error because the `latest` nvidia-mig-parted requires golang 1.23.

Respective error:

```log
go: github.com/NVIDIA/mig-parted/cmd/nvidia-mig-parted@latest: github.com/NVIDIA/mig-parted@v0.12.1 requires go >= 1.23.0 (running go 1.22.2; GOTOOLCHAIN=local)
```

### DIFF
--- a/deployments/systemd/install.sh
+++ b/deployments/systemd/install.sh
@@ -48,7 +48,7 @@ chmod a+rx ${PROFILED_DIR}
 
 ${DOCKER} run --rm \
 	-v ${BINARY_DIR}:/dest \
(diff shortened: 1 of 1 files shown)


elezar:
I have created #212 to ensure that the same go version (or latest) is used for the systemd installer.

PR #215, opened 2025-07-01
Extend the functionality of apply-exit to refresh CDI config

ArangoGutierrez:
This patch adds extra functionality to the `apply-exit` hook, calling the `nvidia-ctk cdi generate` to refresh the CDI config on a MIG (re)configuration event

### DIFF
--- a/deployments/systemd/hooks-default.yaml
+++ b/deployments/systemd/hooks-default.yaml
@@ -16,3 +16,6 @@ hooks:
     - workdir: "/etc/nvidia-mig-manager"
       command: "/bin/bash"
       args: ["-x", "-c", "source hooks.sh; apply-exit"]
+    - workdir: "/etc/nvidia-mig-manager"
+      command: "/bin/bash"
+      args: ["-x", "-c", "source hooks.sh; refresh-cdi"]
--- a/deployments/systemd/hooks.sh
+++ b/deployments/systemd/hooks.sh
@@ -122,3 +122,15 @@ function stop_k8s_pods() {
 	fi
 	return 0
 }
+
+# refresh-cdi triggers the nvidia-cdi-refresh service to regenerate CDI
+# specifications, making updated GPU devices available to container runtimes.
+function refresh-cdi() {
+    # Check if nvidia-cdi-refresh.service exists
+    if systemctl list-unit-files nvidia-cdi-refresh.service --quiet; then
+        echo "Found nvidia-cdi-refresh.service, calling systemctl..." >&2
+        if ! systemctl restart nvidia-cdi-refresh.service; then
+            echo "Error: Failed to start nvidia-cdi-refresh.service" >&2
+        fi
+    fi
+}


deployments/systemd/utils.sh
```diff
@@ -169,3 +169,13 @@ function nvidia-mig-manager::service::kill_k8s_containers_via_containerd_by_imag
...
+
+function nvidia-mig-manager::service::refresh_cdi_config() {
+	local service="nvidia-cdi-refresh.service"
+	systemctl list-unit-files --state=enabled,generated | grep -qF "${service}"
+	if [ "${?}" != "0" ]; then
+		return 0
+	fi
+	systemctl start "${service}"
+	return ${?}
+}
```
klueska: I don't think you should run a service here -- just rely on calling `nvidia-ctk cdi generate` directly

ArangoGutierrez: The thing I am aiming here is to have an unified behaviour, given that with the `nvidia-cdi-refresh` users can modify it's behaviour by adding EV VAR's into a file at `/etc` so I was trying to aim for that. Having just `nvidia-ctk cdi generate` here will not honor the user customizations

deployments/systemd/hooks-default.yaml
```diff
@@ -16,3 +16,7 @@ hooks:
     - workdir: "/etc/nvidia-mig-manager"
       command: "/bin/bash"
       args: ["-x", "-c", "source hooks.sh; apply-exit"]
+    refresh-cdi-config:
```
elezar: The keys in the hooks struct defines point at which one can inject hooks. I suppose it would be sufficient to trigger a CDI refresh as an `apply-exit` hook?

ArangoGutierrez: I thought about it, but an individual hook allows users to opt out. if we call a CDI refresh from the `apply-exit` hook users won't be able to opt out of it.

elezar: Where do we define when the `refresh-cdi-config` hook should be executed?

What I am trying to say is that just defining a new hook "name" in this file doesn't do anything. The hooks are defined here https://github.com/NVIDIA/mig-parted/blob/f04311edd8124463ae30c0620d9ac5fb470a6e98/cmd/nvidia-mig-parted/apply/hooks.go#L23-L27 and have specific names and are triggered at specific points when applying the desired mig config. A user can ALREADY opt-out of any behaviour by updating the hooks spec that is installed.

The question we need to answer is WHEN do we want the CDI specification to be (re)generated. From my perspective `apply-exit` since this is where the MIG configuration would have been changed.

Something like:
```
diff --git a/deployments/systemd/hooks-default.yaml b/deployments/systemd/hooks-default.yaml
index 71ea43eb..f67de092 100644
--- a/deployments/systemd/hooks-default.yaml
+++ b/deployments/systemd/hooks-default.yaml
@@ -16,3 +16,6 @@ hooks:
     - workdir: "/etc/nvidia-mig-manager"
       command: "/bin/bash"
       args: ["-x", "-c", "source hooks.sh; apply-exit"]
+    - workdir: "/etc/nvidia-mig-manager"
+      command: "/bin/bash"
+      args: ["-x", "-c", "nvidia-ctk cdi generate --output=/var/run/cdi/nvidia.yaml"]
```
would probably do what you want it to do -- although, as you point out, we may want to be able to specify environment variables.

ArangoGutierrez: Evan, with your approach, once we merge https://github.com/NVIDIA/nvidia-container-toolkit/pull/1107 maybe that way the user can "just" rely on the toolkit config file, without the need of special ENV VARs.

deployments/systemd/utils.sh
ArangoGutierrez: Edit, this patch now calls `nvidia-ctk cdi generate` directly

deployments/systemd/hooks.sh
```diff
@@ -83,6 +83,14 @@ function apply-exit() {
 	fi
 }
 
+function refresh-cdi-config() {
```
elezar: This is no longer required.

deployments/systemd/hooks-default.yaml
```diff
@@ -16,3 +16,6 @@ hooks:
     - workdir: "/etc/nvidia-mig-manager"
       command: "/bin/bash"
       args: ["-x", "-c", "source hooks.sh; apply-exit"]
+    - workdir: "/etc/nvidia-mig-manager"
+      command: "/bin/bash"
+      args: ["-x", "-c", "nvidia-ctk cdi generate --output=/var/run/cdi/nvidia.yaml"]
```
elezar: Let's add a newline at the end of the file.

ArangoGutierrez: done

deployments/systemd/hooks.sh
ArangoGutierrez: done

deployments/systemd/hooks-default.yaml
elezar: The latest version doesn't address the following:
> would probably do what you want it to do -- although, as you point out, we may want to be able to specify environment variables.

ArangoGutierrez: Given that https://github.com/NVIDIA/nvidia-container-toolkit/pull/1107 is merged, if I remove the `--output` from this PR that should be enough to allow the user to specify custom configuration via the config file

ArangoGutierrez: PR updated

elezar:
@ArangoGutierrez how are we specifying WHERE the output should be created? Although @klueska had concerns about triggering the `nvidia-cdi-refresh.service` doing so would mean that we would ensure consistency in terms of the ouptut path.

Also, on:
> Given that https://github.com/NVIDIA/nvidia-container-toolkit/pull/1107 is merged, if I remove the --output from this PR that should be enough to allow the user to specify custom configuration via the config file

I don't think that specifying an output path in the config file makes sense since this would complicate the default behvaiour which is to output the spec to STDOUT.

ArangoGutierrez:
> @ArangoGutierrez how are we specifying WHERE the output should be created? Although @klueska had concerns about triggering the `nvidia-cdi-refresh.service` doing so would mean that we would ensure consistency in terms of the ouptut path.

I would like to call the `nvidia-cdi-refresh.service` from the hook, maybe something to discuss with @klueska

ArangoGutierrez:
@klueska  / @elezar  Updated based on today's discussion

ArangoGutierrez:
Refresher PING @klueska / @elezar

ArangoGutierrez:
Now that we have Toolkit RC4 and Systemd is stable, can we give this PR another try? 
@klueska / @elezar

deployments/systemd/hooks-default.yaml
```diff
@@ -16,3 +16,7 @@ hooks:
     - workdir: "/etc/nvidia-mig-manager"
       command: "/bin/bash"
       args: ["-x", "-c", "source hooks.sh; apply-exit"]
+    cdi-refresh:
```
elezar: This is not a valid hook name. I think this should probably just be an `apply-exit` hook.

deployments/systemd/cdi-refresh.sh
elezar, line 1: This file is not included in the rpm or deb packages.

elezar (changes requested):
This is **not** what we discussed. We should:

1. Define `cdi-refresh` in `hooks.sh`
2. Add a new hook to the `apply-exit` hooks `hooks-default.yaml` that calls the function in the same way as we do for the other hooks.

We ALSO discussed (not for this PR):
1. Updating the names of the other functions to show intent and not the EVENTS
2. Clarifying the API around envvars etc.

deployments/systemd/hooks.sh
```diff
@@ -122,3 +127,13 @@ function stop_k8s_pods() {
 	fi
 	return 0
 }
+
+cdi-refresh() {
+    # Check if nvidia-cdi-refresh.service exists
+    if systemctl list-unit-files nvidia-cdi-refresh.service --quiet; then
+        echo "Found nvidia-cdi-refresh.service, calling systemctl..."
```
elezar: This should log to STDERR?

ArangoGutierrez: yes, now it logs to STDERR

elezar (dismissed):

deployments/systemd/hooks.sh
```diff
@@ -122,3 +122,13 @@ function stop_k8s_pods() {
...
 	return 0
 }
+
+cdi-refresh() {
+    # Check if nvidia-cdi-refresh.service exists
+    if systemctl list-unit-files nvidia-cdi-refresh.service --quiet; then
+        echo "Found nvidia-cdi-refresh.service, calling systemctl..." >&2
+         if ! systemctl start nvidia-cdi-refresh.service; then
+            echo "Error: Failed to start nvidia-cdi-refresh.service" >&2
+        fi
```
elezar: The indentation is off here.

ArangoGutierrez: fixed

deployments/systemd/hooks.sh
```diff
@@ -122,3 +122,15 @@ function stop_k8s_pods() {
...
 	return 0
 }
+
+# refresh-cdi triggers the nvidia-cdi-refresh service to regenerate CDI
+# specifications, making updated GPU devices available to container runtimes.
+function refresh-cdi() {
+    # Check if nvidia-cdi-refresh.service exists
+    if systemctl list-unit-files nvidia-cdi-refresh.service --quiet; then
+        echo "Found nvidia-cdi-refresh.service, calling systemctl..." >&2
+        if ! systemctl start nvidia-cdi-refresh.service; then
```
elezar: Question: Should this be `restart` instead of `start`?

ArangoGutierrez: You are right, will edit

ArangoGutierrez: Done

PR #216, opened 2025-07-08
Migrate reconfigure to golang

elezar:
This change builds on the proposal from https://github.com/NVIDIA/vgpu-device-manager/pull/93 and migrates the `reconfigure-mig.sh` script to golang.

Ideally I would like to still do the following:
* factor this out so that we can import a package from the `vgpu-device-manager`
* replace the `nvidia-mig-parted` CLI invocations with direct API calls

but these can be done as follow-ups.

### DIFF
--- a/pkg/mig/reconfigure/clients.go
+++ b/pkg/mig/reconfigure/clients.go
@@ -169,9 +169,9 @@ func (o *pod) delete() error {
 	klog.InfoS("Deleting pod", "app", o.app, "node", o.node.name, "namespace", o.namespace)
 	args := []string{
 		"delete", "pod",
-		"--field-selector \"spec.nodeName=" + o.node.name + "\"",
-		"-n \"" + o.namespace + "\"",
-		"-l app=" + o.app,
+		"--field-selector", "spec.nodeName=" + o.node.name,
+		"-n", o.namespace,
+		"-l", "app=" + o.app,
 	}
 
 	klog.InfoS("TODO: running kubctl with args", "args", args)


cmd/nvidia-mig-manager/reconfigure_mig.go
elezar, line 1: This is adapted from https://github.com/NVIDIA/vgpu-device-manager/pull/93

cmd/nvidia-mig-manager/cdi.go
```diff
@@ -0,0 +1,62 @@
...
+		nvcdi.WithVendor("management.nvidia.com"),
+		nvcdi.WithClass("gpu"),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create CDI library for management containers: %v", err)
+	}
+
+	spec, err := cdilib.GetSpec()
+	if err != nil {
+		return fmt.Errorf("failed to genereate CDI spec for management containers: %v", err)
```
tariq1890:
```suggestion
		return fmt.Errorf("failed to genereate CDI spec for management containers: %w", err)
```

cmd/nvidia-mig-manager/cdi.go
```diff
@@ -0,0 +1,62 @@
...
+	cdilib, err := nvcdi.New(
+		nvcdi.WithMode(nvcdi.ModeManagement),
+		nvcdi.WithDriverRoot(opts.DriverRootCtrPath),
+		nvcdi.WithDevRoot(opts.DevRootCtrPath),
+		nvcdi.WithNVIDIACDIHookPath(opts.NVIDIACDIHookPath),
+		nvcdi.WithVendor("management.nvidia.com"),
+		nvcdi.WithClass("gpu"),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create CDI library for management containers: %v", err)
```
tariq1890:
```suggestion
		return fmt.Errorf("failed to create CDI library for management containers: %w", err)
```

cmd/nvidia-mig-manager/cdi.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	transformer := transformroot.NewDriverTransformer(
+		transformroot.WithDriverRoot(opts.DriverRootCtrPath),
+		transformroot.WithTargetDriverRoot(opts.DriverRoot),
+		transformroot.WithDevRoot(opts.DevRootCtrPath),
+		transformroot.WithTargetDevRoot(opts.DevRoot),
+	)
+	if err := transformer.Transform(spec.Raw()); err != nil {
+		return fmt.Errorf("failed to transform driver root in CDI spec: %v", err)
```
tariq1890:
```suggestion
		return fmt.Errorf("failed to transform driver root in CDI spec: %w", err)
```

cmd/nvidia-mig-manager/cdi.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	err = spec.Save("/var/run/cdi/management.nvidia.com-gpu.yaml")
+	if err != nil {
+		return fmt.Errorf("failed to save CDI spec for management containers: %v", err)
```
tariq1890:
```suggestion
		return fmt.Errorf("failed to save CDI spec for management containers: %w", err)
```

cmd/nvidia-mig-manager/clients.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// Stop the specified pod.
+func (o *pod) Stop() error {
+	return o.delete()
+}
+
+func (o *pod) delete() error {
+	err := o.manager.clientset.CoreV1().Pods(o.manager.Namespace).DeleteCollection(
+		context.TODO(),
```
tariq1890: Let's not use `context.TODO()`. Can we pass the application's primary context and modify this method's signature to include `ctx` as a param?

cmd/nvidia-mig-manager/reconfigure_mig.go
```diff
@@ -0,0 +1,452 @@
+/*
```
ArangoGutierrez: Since this file is simply adapted, not much to say besides the Copilot comments

cmd/nvidia-mig-manager/reconfigure_mig.go
```diff
@@ -93,17 +93,9 @@ type reconfigureMIGOptions struct {
 // reboots when necessary. The function ensures that MIG configurations are
 // applied safely with proper service lifecycle management.
 func reconfigureMIG(clientset *kubernetes.Clientset, opts *reconfigureMIGOptions) error {
-	validate := validator.New(validator.WithRequiredStructEnabled())
-
 	log.Info("Validating reconfigure MIG options")
```
ArangoGutierrez: Shouldn't we move this log to the Validate function itself? That way, when called from other locations (if) the logging will also be consistent

cmd/nvidia-mig-manager/clients.go
```diff
@@ -0,0 +1,175 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+)
+
+type k8sClient interface {
+	Restart(*kubernetes.Clientset) error
+}
+
+// An operand is a GPU client that is controlled by a deploy label.
```
ArangoGutierrez: It was my understanding that an Operand is a deployment managed/owned by a controller (operator)

cmd/nvidia-mig-manager/clients.go
```diff
@@ -0,0 +1,175 @@
...
+type operand struct {
+	app         string
+	deployLabel string
+	lastValue   string
+}
+
+func stopK8sClients(clientset *kubernetes.Clientset) ([]k8sClient, error) {
+	// TODO: We need to add this namespace to the options.
```
ArangoGutierrez: ++

cmd/nvidia-mig-manager/clients.go
```diff
@@ -0,0 +1,175 @@
...
+	gpuClientNamespace := "nvidia-gpu-operator"
+
+	var k8sGPUClients []k8sClient
+
+	// We first optionally stop the operands managed by the operator:
+	var operands = []*operand{
```
ArangoGutierrez: For now hard coded is ok, but should we query the API server filtered with label "app=gpu-operator" and grab all the current operands managed by the operator? that way we don't have to manage this list

cmd/nvidia-mig-manager/reconfigure_mig.go
```diff
@@ -61,23 +63,26 @@ type reconfigureMIGOptions struct {
...
 	// HostRootMount is the container path where host root directory is mounted.
 	HostRootMount string `validate:"dirpath"`
 
 	// HostMIGManagerStateFile is the path where the systemd mig-manager state
 	// file is located.
 	HostMIGManagerStateFile string `validate:"filepath"`
 
+	// WithShutdownHostGPUClients shutdowns/restarts any required host GPU clients
+	// across a MIG configuration.
+	WithShutdownHostGPUClients bool
```
ArangoGutierrez: are we moving this down due to alphabetical order? if so can we push it all the way down, same with `WithReboot` 
just a NIT, not a blocker

cmd/nvidia-mig-manager/cdi.go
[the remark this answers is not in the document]
ArangoGutierrez: ++

cmd/nvidia-mig-manager/cdi.go
```diff
@@ -0,0 +1,62 @@
...
+**/
+
+package main
+
+import (
+	"fmt"
+
+	"github.com/NVIDIA/nvidia-container-toolkit/pkg/nvcdi"
+	transformroot "github.com/NVIDIA/nvidia-container-toolkit/pkg/nvcdi/transform/root"
+	log "github.com/sirupsen/logrus"
```
ArangoGutierrez: I'm surprised golangci-lint is not complaining about this, but could we add a new line between the NVIDIA packages and the non-NVIDIA ones

cmd/nvidia-mig-manager/cdi.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+func regenerateCDISpec(opts *reconfigureMIGOptions) error {
+	log.Info("Generating CDI spec for management containers")
+	cdilib, err := nvcdi.New(
```
ArangoGutierrez: the new API 😍

cmd/nvidia-mig-manager/reconfigure_mig.go
```diff
@@ -0,0 +1,531 @@
...
+	MIGPartedConfigFile string `validate:"required,filepath"`
+
+	// SelectedMIGConfig is the selected mig-parted configuration to apply to the
+	// node.
+	SelectedMIGConfig string
+
+	// ConfigStateLabel is the label that is used to indicate the state of the config being applied.
+	ConfigStateLabel string
+
+	// DriverLibrayPath is the path to libnvidia-ml.so.1 in the container.
```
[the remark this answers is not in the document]
ArangoGutierrez: ++

cmd/nvidia-mig-manager/main.go
```diff
@@ -112,17 +113,23 @@ func (m *SyncableMigConfig) Get() string {
 }
 
 func main() {
+	o := &options{}
+
 	c := cli.NewApp()
```
ArangoGutierrez: Note to self: this is a candidate for V3 migration

cmd/nvidia-mig-manager/clients.go
```diff
@@ -26,9 +25,9 @@ type withNoRestartClient struct {
 }
 
 type pod struct {
-	namespace string
-	nodename  string
-	app       string
+	app string
+	// manager stores a reference to the mig manager managing these clients.
```
ArangoGutierrez: "A manager stores a reference to the manager handling these clients."

cmd/nvidia-mig-manager/clients.go
```diff
@@ -9,89 +9,78 @@ import (
...
 type withNoStopClient struct {
 	k8sClient
 }
 
+// a withNoRestartClient wraps the specified client so that the Restart function is a no-op.
 type withNoRestartClient struct {
 	k8sClient
 }
 
+// A pod represents a kubernetes pod with a specified app= label.
```
ArangoGutierrez: Note that in the type above, the "a" is lower case; we should do all lower or all upper, for consistency

ArangoGutierrez (commented):
First pass, looks good

jgehrcke:
> migrates the reconfigure_mig.sh script to golang.

Where can I find `reconfigure_mig.sh`?

ArangoGutierrez:
> > migrates the reconfigure_mig.sh script to golang.
> 
> Where can I find `reconfigure_mig.sh`?

https://github.com/NVIDIA/mig-parted/blob/main/deployments/container/reconfigure-mig.sh

cmd/nvidia-mig-manager/cdi.go
tariq1890: Why should it? All of these imports are non-local and third party; they should be grouped together

The accepted import grouping is:
```
stdlib imports

third-party imports

local imports
```

cmd/nvidia-mig-manager/reconfigure_mig.go
```diff
@@ -0,0 +1,531 @@
...
+		return false
+	}
+
+	return systemdServicePrefixPattern.MatchString(serviceName)
+}
+
+func (m *migManager) getNodeLabelValue(label string) (string, error) {
+	node, err := m.clientset.CoreV1().Nodes().Get(context.TODO(), m.NodeName, metav1.GetOptions{})
+	if err != nil {
+		return "", fmt.Errorf("unable to get node object: %v", err)
```
tariq1890:
```suggestion
		return "", fmt.Errorf("unable to get node object: %w", err)
```

cmd/nvidia-mig-manager/reconfigure_mig.go
```diff
@@ -0,0 +1,531 @@
...
+
+	if !hasSuffix {
+		return false
+	}
+
+	return systemdServicePrefixPattern.MatchString(serviceName)
+}
+
+func (m *migManager) getNodeLabelValue(label string) (string, error) {
+	node, err := m.clientset.CoreV1().Nodes().Get(context.TODO(), m.NodeName, metav1.GetOptions{})
```
tariq1890: Let's propagate `ctx` down the method call stack

cmd/nvidia-mig-manager/reconfigure_mig.go
```diff
@@ -0,0 +1,531 @@
...
+		return "", nil
+	}
+
+	return value, nil
+}
+
+func (m *migManager) setNodeLabelValue(label, value string) error {
+	node, err := m.clientset.CoreV1().Nodes().Get(context.TODO(), m.NodeName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to get node object: %v", err)
```
tariq1890:
```suggestion
		return fmt.Errorf("unable to get node object: %w", err)
```

cmd/nvidia-mig-manager/reconfigure_mig.go
```diff
@@ -0,0 +1,531 @@
...
+	}
+
+	labels := node.GetLabels()
+	labels[label] = value
+	node.SetLabels(labels)
+	_, err = m.clientset.CoreV1().Nodes().Update(context.TODO(), node, metav1.UpdateOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to update node object: %v", err)
```
tariq1890:
```suggestion
		return fmt.Errorf("unable to update node object: %w", err)
```

cmd/nvidia-mig-manager/clients.go
elezar: Sure, but the relevant information here is that it is something that uses the GPU. This is why it needs to be shut down.

cmd/nvidia-mig-manager/clients.go
elezar: For the first iteration I wanted to mirror what is being done in `reconfigure_mig.sh`. Once we have a stable go-based implementation we have much more flexibility.

Also note that the mig manager itself should NOT be stopped in this was as it is busy performing the required updated. The NVIDIA Container Toolkit container is also NOT a GPU client and doesn't need to be stopped.

cmd/nvidia-mig-manager/cdi.go
elezar: Yes, the imports from the `nvidia-container-toolkit` are non-local i.e. third-party -- although they are also NVIDIA-owned.

cmd/nvidia-mig-manager/cdi.go
elezar: This isn't the new API. This API has existed for as long as the `nvcdi` package has existed. The "New" API was just cleaning up some internals and REMOVING functions that were largely unused.

jgehrcke:
> migrates the reconfigure_mig.sh script to golang.

About my: "Where can I find reconfigure_mig.sh?" Thanks @ArangoGutierrez -- I was searching quite wildly for `reconfigure_mig.sh`, but this was an underscore vs. hyphen problem.

Permanent URL: https://github.com/NVIDIA/mig-parted/blob/15f09c791454833c44a78f5b44a1bcd6a62ec6d0/deployments/container/reconfigure-mig.sh

cmd/nvidia-mig-manager/main.go
```diff
@@ -377,30 +377,41 @@ func runScript(migConfigValue string, driverLibraryPath string, nvidiaSMIPath st
...
+		reconfigure.WithHostKubeletService(hostKubeletSystemdServiceFlag),
+		reconfigure.WithGPUClientNamespace(defaultGPUClientsNamespaceFlag),
 	}
+
 	if cdiEnabledFlag {
-		args = append(args, "-e", "-t", driverRoot, "-a", driverRootCtrPath, "-b", devRoot, "-j", devRootCtrPath, "-l", driverLibraryPath, "-q", nvidiaSMIPath, "-s", nvidiaCDIHookPath)
-	}
-	if withRebootFlag {
-		args = append(args, "-r")
+		options = append(options,
```
cdesiniotis, line 398: Question -- do we actually need the conditional here? Or can we always add these options to the options slice?

cmd/nvidia-mig-manager/main.go
```diff
@@ -377,30 +377,41 @@ func runScript(migConfigValue string, driverLibraryPath string, nvidiaSMIPath st
...
+			reconfigure.WithDevRootCtrPath(devRootCtrPath),
+			reconfigure.WithDriverLibraryPath(driverLibraryPath),
+			reconfigure.WithNVIDIASMIPath(nvidiaSMIPath),
+			reconfigure.WithNVIDIACDIHookPath(nvidiaCDIHookPath),
+		)
 	}
-	if withShutdownHostGPUClientsFlag {
-		args = append(args, "-d")
+
+	options = append(options,
```
cdesiniotis, line 410: Question -- is there any value in appending these options here? Can we add them to the list of options during initialization of the options slice?

pkg/mig/reconfigure/api.go
cdesiniotis, line 1: nit: this file is missing a license header

pkg/mig/reconfigure/options.go
```diff
@@ -0,0 +1,204 @@
...
+	// DevRoot is the path where the NVIDIA device nodes are found on in the container.
+	DevRootCtrPath string
+
+	// DriverRoot is the path to the NVIDIA driver installation on the HOST.
+	DriverRoot string
+	// DevRoot is the path where the required device nodes are created on the HOST.
+	DevRoot string
+
+	CDIEnabled    bool
+	NVIDIASMIPath string
```
cdesiniotis, line 89: nit: let's add a comment for this field indicating whether the path is relative to the container or host

pkg/mig/reconfigure/options.go
```diff
@@ -0,0 +1,204 @@
...
+	// NVIDIACDIHookPath is the path to the nvidia-cdi-hook executable on the HOST.
+	NVIDIACDIHookPath string
+
+	hostNVIDIADir string
```
cdesiniotis, line 93: nit: let's add a comment documenting this field

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,513 @@
...
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+const (
+	configStatePending   = "pending"
+	configStateRebooting = "rebooting"
```
cdesiniotis, line 39: Question -- shouldn't the "failed" and "success" states be defined as constants as well?

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,513 @@
...
+
+	log.Infof("Shutting down all GPU clients in Kubernetes by disabling their component-specific nodeSelector labels")
+	if err := k8sGPUClients.Stop(); err != nil {
+		// TODO: Update this error message.
+		return fmt.Errorf("unable to tear down GPU client pods by setting their daemonset labels: %w", err)
+	}
+
+	if opts.WithShutdownHostGPUClients {
+		log.Info("Shutting down all GPU clients on the host by stopping their systemd services")
+		if err := opts.hostStopSystemdServices(systemdClients); err != nil {
```
cdesiniotis, line 272: This doesn't appear right to me. `systemdClients` is declared at the top of this function (line 177) but it is never populated with values. Shouldn't we be stopping all the services present in the `opts.HostGPUClientServices` slice?

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,513 @@
...
+
+		// Ensure that we regenerate a CDI spec for management containers.
+		if err := opts.regenerateManagementCDISpec(); err != nil {
+			return err
+		}
+	}
+
+	if opts.WithShutdownHostGPUClients {
+		log.Info("Restarting all GPU clients previously shutdown on the host by restarting their systemd services")
+		if err := opts.hostStartSystemdServices(systemdClients); err != nil {
```
cdesiniotis, line 321: Same comment as before. I don't see `systemdClients` being set to anything.

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,513 @@
...
+func (opts *reconfigureMIGOptions) hostStopSystemdServices(systemdGPUClients gpuClients) error {
+	for _, serviceName := range opts.HostGPUClientServices {
+		service := opts.newSystemdService(serviceName)
+
+		mustRestart, err := service.Pause()
+		if err != nil {
+			return err
+		}
+		if mustRestart {
+			systemdGPUClients = append(systemdGPUClients, service)
```
cdesiniotis, line 431: Okay so we are appending to this list here... this is hard to follow. I just read the original bash script and I better understand what this is doing now. Some of my previous comments / questions can be disregarded now.

Question -- should we be passing the `systemGPUClients` slice by reference instead of by value? I believe the answer is yes. When we `append()` here, a new slice gets created and the original `systemdGPUClients` identifier at the call-site will not see the newly created slice.

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,513 @@
...
+			}
+		}
+	}
+
+	// TODO: We should allow restarts to continue on failure.
+	if err := systemdGPUClients.Restart(); err != nil {
+		return fmt.Errorf("some services failed to start: %w", err)
+	}
+	return nil
+}
```
cdesiniotis, line 453: Question -- instead of maintaining a list of `systemdGPUClients` that need to be restarted BEFORE calling this function, what if we always construct this list in this function? That would eliminate the need to pass any arguments to `hostStopSystemdServices()` / `hostStartSystemdServices()` and make this much more readable in my opinion.

This is potentially out-of-scope for this PR which is simply a 1:1 port of the bash script to Go.

cdesiniotis (commented):
First pass. I have not reviewed everything in detail yet.

pkg/mig/reconfigure/api.go
```diff
@@ -0,0 +1,41 @@
...
+	assertMIGConfig() error
+	assertMIGModeOnly() error
+	applyMIGModeOnly() error
+	applyMIGConfig() error
+}
+
+// nodeLabeller defines an interface for interacting with node labels.
+//
+//go:generate moq -rm -fmt=goimports -out node-labeller_mock.go . nodeLabeller
+type nodeLabeller interface {
```
tariq1890, line 37: Can we approach the interface differently? What if we had an interface that captured all the Kubernetes APIServer actions? The interface would have methods where each one is a wrapper over kube client calls

pkg/mig/reconfigure/cdi.go
```diff
@@ -0,0 +1,87 @@
...
+		nvcdi.WithVendor("management.nvidia.com"),
+		nvcdi.WithClass("gpu"),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create CDI library for management containers: %v", err)
+	}
+
+	spec, err := cdilib.GetSpec()
+	if err != nil {
+		return fmt.Errorf("failed to genereate CDI spec for management containers: %v", err)
```
tariq1890, line 45:
```suggestion
		return fmt.Errorf("failed to generate CDI spec for management containers: %w", err)
```

pkg/mig/reconfigure/cdi.go
```diff
@@ -0,0 +1,87 @@
...
+	cdilib, err := nvcdi.New(
+		nvcdi.WithMode(nvcdi.ModeManagement),
+		nvcdi.WithDriverRoot(opts.DriverRootCtrPath),
+		nvcdi.WithDevRoot(opts.DevRootCtrPath),
+		nvcdi.WithNVIDIACDIHookPath(opts.NVIDIACDIHookPath),
+		nvcdi.WithVendor("management.nvidia.com"),
+		nvcdi.WithClass("gpu"),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create CDI library for management containers: %v", err)
```
tariq1890, line 40:
```suggestion
		return fmt.Errorf("failed to create CDI library for management containers: %w", err)
```

pkg/mig/reconfigure/cdi.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	transformer := transformroot.NewDriverTransformer(
+		transformroot.WithDriverRoot(opts.DriverRootCtrPath),
+		transformroot.WithTargetDriverRoot(opts.DriverRoot),
+		transformroot.WithDevRoot(opts.DevRootCtrPath),
+		transformroot.WithTargetDevRoot(opts.DevRoot),
+	)
+	if err := transformer.Transform(spec.Raw()); err != nil {
+		return fmt.Errorf("failed to transform driver root in CDI spec: %v", err)
```
tariq1890, line 55:
```suggestion
		return fmt.Errorf("failed to transform driver root in CDI spec: %w", err)
```

pkg/mig/reconfigure/cdi.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	err = spec.Save("/var/run/cdi/management.nvidia.com-gpu.yaml")
```
tariq1890, line 57: can we move `"/var/run/cdi"` to a constant?

pkg/mig/reconfigure/cdi.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to save CDI spec for management containers: %v", err)
```
tariq1890, line 59:
```suggestion
		return fmt.Errorf("failed to save CDI spec for management containers: %w", err)
```

pkg/mig/reconfigure/cdi.go
```diff
@@ -0,0 +1,87 @@
...
+		"system", "create-device-nodes",
+		"--control-devices",
+		"--dev-root=" + opts.DevRootCtrPath,
+	}
+	cmd := exec.Command("nvidia-ctk", args...)
+
+	return opts.Run(cmd)
+}
+
+func (opts *reconfigurer) runNvidiaSMI() error {
```
tariq1890, line 78: Instead of running `nvidia-smi`, can we use `go-nvlib/go-nvml` to retrieve the relevant info?

pkg/mig/reconfigure/clients.go
```diff
@@ -0,0 +1,344 @@
...
+	"os/exec"
+	"strings"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+// A gpuClient represents a client that can be stoped or restarted.
```
tariq1890, line 35:
```suggestion
// A gpuClient represents a client that can be stopped or restarted.
```

pkg/mig/reconfigure/clients.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+// Stop the specified pod.
+func (o *pod) Stop() error {
+	return o.delete()
+}
+
+func (o *pod) delete() error {
+	err := o.node.clientset.CoreV1().Pods(o.namespace).DeleteCollection(
+		context.TODO(),
```
tariq1890: we should propagate the cli context here

pkg/mig/reconfigure/clients.go
```diff
@@ -0,0 +1,344 @@
...
+	)
+	if err != nil {
+		return fmt.Errorf("unable to delete pods for app %s: %w", o.app, err)
+	}
+	return nil
+}
+
+func (o *pod) waitForDeletion() error {
+	timeout := 5 * time.Minute
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
tariq1890, line 201: propagate CLI context here

pkg/mig/reconfigure/clients.go
```diff
@@ -0,0 +1,344 @@
...
+
+func (opts *reconfigureMIGOptions) newSystemdService(name string) *systemdService {
+	return &systemdService{
+		name:          name,
+		hostRootMount: opts.HostRootMount,
+	}
+}
+
+func (o *systemdService) Restart() error {
+	cmd := exec.Command("chroot", o.hostRootMount, "systemctl", "start", o.name) // #nosec G204 -- HostRootMount validated via dirpath, service validated via systemd_service_name.
```
tariq1890, line 299: Can we use a go library instead of shelling out?

Would this work? https://github.com/coreos/go-systemd

rajathagasthya:
Similar to #218, closing this one as an alternate implementation was merged in #244

PR #240, opened 2025-09-05
Updated NGC Publishing Configs and NSPECT changes

JunAr7112:


.nvidia-ci.yml
```diff
@@ -105,3 +105,89 @@ image-ubuntu20.04:
...
+# Create a dummy MR that exercises the publishing logic
+mock-image-publish:
+  extends:
+    - .publish-images
+  except:
+    - tags
+  needs:
+    - job: update-nspect-staging
+  variables:
+    NGC_PUBLISHING_PROJECT_PATH: "${NGC_PUBLISHING_TEST_PROJECT_PATH}"
```
tariq1890: add a new line

.nvidia-ci.yml
```diff
@@ -105,3 +105,90 @@ image-ubuntu20.04:
...
+
```
tariq1890: Looks like it is still unresolved

JunAr7112: There seems to be a new line (194)

cdesiniotis (approved):

.nvidia-ci.yml
```diff
@@ -105,3 +105,89 @@ image-ubuntu20.04:
...
+    - job: release:staging-ubi9
+    - job: release:staging-ubuntu20.04
+  extends:
+    - .ngc-publish-variables
+  image: 
+    name: "${CNT_NGC_PUBLISH_IMAGE}"
+    pull_policy: always
+  variables:
+    PROJECT_NAME: "k8s-mig-manager"
+    REPO_URL: "https://github.com/NVIDIA/mig-parted.git"
```
elezar, line 130: Not critical for this PR, but more of a question: for the scripting to create the publication MRs, a lot of this information is defined in the tooling itself. How do we determine which information is defined at the project-level and which is defined at the tooling level?

.nvidia-ci.yml
```diff
@@ -105,3 +105,89 @@ image-ubuntu20.04:
...
+  needs:
+    - job: release:staging-ubi9
+    - job: release:staging-ubuntu20.04
+  extends:
+    - .ngc-publish-variables
+  image: 
+    name: "${CNT_NGC_PUBLISH_IMAGE}"
+    pull_policy: always
+  variables:
+    PROJECT_NAME: "k8s-mig-manager"
```
elezar: NIT (out of scope): Why is the project name in nspect and for the `cnt-ngc-publish render` script not consistent?

Also, why do we specify this as an ENVVAR here, but on the command line when we call `cnt-ngc-publish render`?

JunAr7112: For some of the projects like mig-parted the image name we are registering ie REGISTRY/k8s-mig-manager differs from the project name (mig-parted). That's the reason why the PROJECT_NAME being used to register in NSPECT differs from the project reference used in cnt-ngc-publish render

.nvidia-ci.yml
JunAr7112: Usually we try to keep as much of the actual publishing logic in the tooling scripts which are accessed through cnt-ngc-publish. We try to limit the information at the project-level to either info that varies by project (ie the project_name, source repo, OSRB ID) or info that will depend on the release job type (staging vs production credentials, release version)

.nvidia-ci.yml
JunAr7112: Following up on this discussion, I will consistently set the PROJECT_NAME in both the Publish and NSPECT jobs to be mig-parted and update the cnt-ngc-publish command accordingly. This should make the formatting clearer.

PR #244, opened 2025-09-24
port reconfigure-mig.sh to go

tariq1890:


deployments/container/Dockerfile.ubi9
```diff
@@ -45,29 +45,21 @@ ARG VERSION="N/A"
 ARG GIT_COMMIT="unknown"
 RUN make PREFIX=/artifacts cmds
 
-RUN cp ./deployments/container/reconfigure-mig.sh /artifacts/reconfigure-mig.sh
```
cdesiniotis, line 48: I don't see the reconfigure.sh file removed as part of this PR. Is that intended?

internal/systemd/systemd.go
```diff
@@ -0,0 +1,270 @@
...
+			Type:     "",
+			SubState: "not-found",
+		}, nil
+	}
+
+	// Get unit properties
+	properties, err := sm.conn.GetAllPropertiesContext(sm.ctx, serviceName)
+	if err != nil {
+		// If the unit doesn't exist, return a status indicating it doesn't exist
+		if strings.Contains(err.Error(), "not found") {
```
cdesiniotis, line 90: I'm assuming the dbus package does not define any custom error types that we could explicitly check for instead?

internal/systemd/systemd_test.go
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package systemd
+
+import (
+	"testing"
+)
+
+func TestServiceStatus(t *testing.T) {
```
cdesiniotis: I am not sure I see what value this test function currently brings. It is not really testing anything.

internal/systemd/systemd.go
```diff
@@ -0,0 +1,270 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
```
cdesiniotis: nit: copyright year

internal/systemd/systemd_test.go
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
```
cdesiniotis: nit: copyright year

pkg/mig/reconfigure/reconfigure_test.go
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
```
cdesiniotis: nit: copyright year

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,674 @@
...
+		hostSystemBusAddress := fmt.Sprintf("unix:path=%s/run/dbus/system_bus_socket", opts.HostRootMount)
+		_ = os.Setenv("DBUS_SYSTEM_BUS_ADDRESS", hostSystemBusAddress)
+	}
+
+	// Initialize systemd managers
+	// Try to connect to systemd D-Bus for local operations
+	systemdManager, err := systemd.NewManager(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to initialize systemd manager: %w", err)
+	}
```
cdesiniotis: A couple of comments:
1) The comment indicates a systemdManager is only created under certain conditions, but it appears to be constructed unconditionally.
2) Is creating a systemdManager when we have a host root mount the right conditional? For mig-manager, we always mount the host root regardless if drivers are pre-installed on the host or not. Would constructing a systemdManager only when `WITH_SHUTDOWN_HOST_GPU_CLIENT` is `true` be the "right" conditional?

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,674 @@
...
+		systemdManager: systemdManager,
+	}, nil
+}
+
+// Run executes the complete MIG reconfiguration process
+func (r *Reconfigure) Run() error {
+	fmt.Println("Starting MIG reconfiguration process...")
+
+	// Ensure systemd managers are cleaned up
+	defer r.cleanup()
```
cdesiniotis, line 124: Question: what does this do exactly?

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,674 @@
...
+
+	// Step 1: Get current node labels
```
cdesiniotis: nit: I don't find most of these comments particularly useful since the method names themselves make it clear what each step is.

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,674 @@
...
+	}
+
+	// Step 4: Persist config if needed
+	if err := r.persistConfigIfNeeded(); err != nil {
+		return fmt.Errorf("failed to persist config: %w", err)
+	}
+
+	// Step 5: Check if MIG mode change is required
+	if err := r.checkMigModeChangeRequired(); err != nil {
+		return fmt.Errorf("failed to check MIG mode change requirement: %w", err)
```
cdesiniotis: nit:
```suggestion
		return fmt.Errorf("failed to check if a MIG mode change is required: %w", err)
```

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,674 @@
...
+	}
+
+	// Step 6: Set state to pending
+	if err := r.setNodeLabel("nvidia.com/mig.config.state", "pending"); err != nil {
```
cdesiniotis: Can we move the MIG config state label to a constant?

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,674 @@
...
+	fmt.Printf("Current value of 'nvidia.com/gpu.deploy.nvsm=%s'\n", r.nvsmDeployed)
+	fmt.Printf("Current value of 'nvidia.com/mig.config.state=%s'\n", r.currentState)
+
+	return nil
+}
+
+// validateMigConfig validates the MIG configuration
+func (r *Reconfigure) validateMigConfig() error {
+	fmt.Println("Asserting that the requested configuration is present in the configuration file")
+	// TODO: need to invoke this method via Go
```
cdesiniotis, line 275: Thanks for adding this TODO. I would not say this is a P0, so this can be done in a follow-up.

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,674 @@
...
+	}
+
+	return r.setNodeLabels(labels)
+}
+
+// waitForPodsToBeDeleted waits for GPU client pods to be deleted
+func (r *Reconfigure) waitForPodsToBeDeleted() error {
+	timeout := 5 * time.Minute
+
+	// Wait for device-plugin pod
```
cdesiniotis: nit: I don't find this comment (and the others in this function) particularly useful.

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,674 @@
...
+	}
+
+	return nil
+}
+
+// shutdownHostGPUClients shuts down host GPU clients
+func (r *Reconfigure) shutdownHostGPUClients() error {
+	fmt.Println("Shutting down all GPU clients on the host by stopping their systemd services")
+
+	// Copy nvidia-mig-parted to host
```
cdesiniotis: nit: I don't find this comment useful

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,674 @@
...
+	if err := r.setNodeLabel("nvidia.com/mig.config.state", state); err != nil {
+		fmt.Printf("Unable to set 'nvidia.com/mig.config.state' to '%s'\n", state)
+		if state != "success" {
+			fmt.Println("Exiting with incorrect value in 'nvidia.com/mig.config.state'")
+		}
+		return err
+	}
+
+	if exitCode != 0 {
+		os.Exit(exitCode)
```
cdesiniotis: Is exiting actually the behavior we want? This will cause the pod to crash, which does not align with the behavior prior to this PR IIUC.

Before this PR, the `nvidia-mig-manager` Go binary would shell out and invoke the `reconfigure_mig.sh` script -- it would use the exit code of the `reconfigure_mig.sh` script as its "return" value. Now that we rewrote the `reconfigure_mig.sh` script in Go, we should be returning errors in favor of throwing exit codes.

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -0,0 +1,674 @@
...
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+func (r *Reconfigure) createCDISpec() error {
+	// This is a simplified version of the complex CDI spec creation
+	// In a real implementation, you would need to implement the full pipeline
+	fmt.Println("Creating management CDI spec (simplified implementation)")
+	return nil
+}
```
cdesiniotis, line 699: This is not implemented.

cdesiniotis (commented):
@tariq1890 just made a first pass -- this was really easy to read, great work!

deployments/container/Dockerfile.ubi9
tariq1890: Yes, that is intended. I thought it'd be used to have it around for a little longer. If we were to run into any regresssion, we could look back to the script

internal/systemd/systemd.go
tariq1890: Will update

internal/systemd/systemd.go
tariq1890: Unfortunately, this error block will most likely never be executed. What I've observed that this method just returns an empty properties map when you provide a bogus systemd unit/service as input

internal/systemd/systemd_test.go
tariq1890: Will update

internal/systemd/systemd_test.go
tariq1890: You're right. This is Cursor/LLM slop that I need to remove

pkg/mig/reconfigure/reconfigure.go
tariq1890: Will do

pkg/mig/reconfigure/reconfigure.go
tariq1890: Will do

pkg/mig/reconfigure/reconfigure.go
tariq1890: Will remove

pkg/mig/reconfigure/reconfigure.go
tariq1890: 1. I'll remove that comment. The systemdManager is to be created unconditionally
2. We create the systemdManager always. When the hostRootMount is provided, we just set the `DBUS_SYSTEM_BUS_ADDRESS` env var.

pkg/mig/reconfigure/reconfigure.go
tariq1890: let me double-check this

pkg/mig/reconfigure/reconfigure.go
tariq1890: this is a method stub that needs to be remove. We invoke the `nvidia-cdi-hook` command elsewhere

tariq1890: whoops, it looks like I completely missed filling this method out. Will address this :)

pkg/mig/reconfigure/reconfigure.go
tariq1890: The systemdManager is initialised by opening a connection with DBUS socket. This has to be closed when the programme execution terminates

pkg/mig/reconfigure/reconfigure.go
tariq1890: Agreed. I will address this

pkg/mig/reconfigure/reconfigure_test.go
tariq1890: Updated

pkg/mig/reconfigure/reconfigure.go
tariq1890: removed

pkg/mig/reconfigure/reconfigure.go
tariq1890: I've addressed this. Please check again and let me know

pkg/mig/reconfigure/reconfigure.go
tariq1890: This is done. Please review

tariq1890:
@cdesiniotis Ready for another round of review

internal/systemd/systemd.go
tariq1890: I'll remove the above comment as it is misleading. It is llm-generated

tariq1890: comment removed

pkg/mig/reconfigure/reconfigure.go
cdesiniotis: I still see these comments throughout this function.

tariq1890: Sure, I can remove

tariq1890: Done

cdesiniotis: Thanks.

PR #246, opened 2025-09-25
add support for building images with custom GOPROXY

tariq1890:


deployments/container/Dockerfile.ubuntu
cdesiniotis, line 1: Was this file added by mistake?

tariq1890: Good catch. I thought GoLand would delete this file after resolve the merge conflict

tariq1890: Fixed

PR #259, opened 2025-10-15
create symlink from /var/run to /run

tariq1890:
This change is needed for the nvidia-smi command to be able to detect the nvidia-persistenced socket. The container toolkit when injecting GPU driver components to a container mounts the persistenced socket at `/run/nvidia-persistenced/socket`. Creating this symlink will allow nvidia-smi
to reach the actual socket path and perform GPU resets successfully.

### DIFF
--- a/deployments/container/Dockerfile
+++ b/deployments/container/Dockerfile
@@ -59,7 +59,11 @@ FROM nvcr.io/nvidia/distroless/go:v3.1.13-dev
 
 USER 0:0
 SHELL ["/busybox/sh", "-c"]
-RUN ln -s /busybox/sh /bin/sh
+RUN ln -s /busybox/sh /bin/sh && \
+    # delete /var/run as it's an empty dir
+    rm -r /var/run && \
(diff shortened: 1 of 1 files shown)


deployments/container/Dockerfile
```diff
@@ -59,7 +59,8 @@ FROM nvcr.io/nvidia/distroless/go:v3.1.13-dev
 
 USER 0:0
 SHELL ["/busybox/sh", "-c"]
-RUN ln -s /busybox/sh /bin/sh
+RUN ln -s /busybox/sh /bin/sh && \
+    ln -s /run /var/run
```
cdesiniotis, line 66: Since `/var/run` already exists, I believe you need to delete it first.
```suggestion
    rm -r /var/run && ln -s /run /var/run
```

tariq1890: Good catch, thanks!

tariq1890: Updated, please check now

PR #261, opened 2025-10-21
Migrate to nvcdi api

elezar:
Ideally we would also switch away from using the nvidia-ctk CLI to create device nodes as part of this PR too.


pkg/mig/reconfigure/reconfigure.go
```diff
@@ -468,11 +473,13 @@ func (r *Reconfigure) applyMigConfig() error {
...
-	// TODO: Instead of shelling out, we need to invoke the method via Go. The Toolkit code needs to be refactored first.
-	cmd := exec.Command("nvidia-ctk", "system", "create-device-nodes", "--control-devices", "--dev-root="+r.opts.DevRootCtrPath)
-	cmd.Stdout = os.Stdout
-	cmd.Stderr = os.Stderr
-	if err := cmd.Run(); err != nil {
+
+	// TODO: Instead of abusing CLI command we generate here, we should expose
+	// this API and use that instead. This would require refactoring in the
+	// toolkit.
+	cmd := devicenodes.NewCommand(log.StandardLogger())
```
elezar, line 482: See #262 for a follow-up.

cmd/nvidia-mig-manager/main.go
```diff
@@ -418,7 +418,7 @@ func migReconfigure(ctx context.Context, migConfigValue string, clientset *kuber
 		return fmt.Errorf("error creating reconfigure instance: %w", err)
 	}
 
-	return rcfg.Run()
+	return rcfg.Run(ctx)
```
tariq1890: we already `ctx` in the `rcfg, err := reconfigure.New(ctx, clientset, migPartedBinary, opts)` line. Can we access the context from the struct itself?

elezar: Yes. You're right. I can use the context member.

elezar: Updated.

rajathagasthya:
New PR will be created when #362 is worked on.

PR #262, opened 2025-10-21
Use nvdevices API

elezar:
Requires https://github.com/NVIDIA/nvidia-container-toolkit/pull/1372


rajathagasthya:
New PR will be created when #361  is worked on.

PR #269, opened 2025-10-29
Build multi arch packages on native GitHub runners

rajathagasthya:
Like how we build images on native runners, this change builds packages (dep, rpm, tarballs) on native runners.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -23,9 +23,11 @@ on:
 
 jobs:
   packages:
-    runs-on: ubuntu-latest
     strategy:
       matrix:
+        arch:
+          - amd64
+          - arm64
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -20,19 +20,23 @@ on:
     branches:
       - main
       - release-*
+      - "pull-request/[0-9]+"
```
rajathagasthya: Should we enable this job for PRs too?

tariq1890: I'd rather this be a separate change and we can go over the pros/cons enabling package builds in PR pipelines in a separate PR thread

tariq1890 (approved):

PR #275, opened 2025-11-10
Use functional names for hook bash functions

ArangoGutierrez:
Rename hook functions to describe their actions rather than when they run. Add documentation comments for clarity. No API changes.

### DIFF
--- a/deployments/systemd/hooks.sh
+++ b/deployments/systemd/hooks.sh
@@ -46,28 +46,9 @@ function persist-config() {
 	return 0
 }
 
-# stop-services-for-mode stops all GPU-dependent services (K8s workloads
-# and driver services) before applying MIG mode changes that require a
-# GPU reset.
-function stop-services-for-mode() {
-	stop_k8s_services
-	if [ "${?}" != "0" ]; then
-		return 1
-	fi
-	stop_k8s_pods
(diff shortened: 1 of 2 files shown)


deployments/systemd/hooks.sh
```diff
@@ -44,7 +46,10 @@ function apply-start() {
 	return 0
 }
 
-function pre-apply-mode() {
+# stop-services-for-mode stops all GPU-dependent services (K8s workloads
+# and driver services) before applying MIG mode changes that require a
+# GPU reset.
+function stop-services-for-mode() {
```
elezar: Instead of duplicating the `stop_k8s_services` and `stop_k8s_pods` across `stop-services-for-mode` and `stop-services-for-config`, should we invoke `stop_driver_services` directly as a hook from `pre-apply-mode` and invoke `stop-services` from both `apply-mode` and `pre-apply-config`?

ArangoGutierrez: Check : [Refactor: move service stop orchestration to hook configuration level](https://github.com/NVIDIA/mig-parted/pull/275/commits/aeca231f01cd56a2da8a02c431ce9bf564cebe8d) , added as an additional commit. If you approve, I'lll edit the PR description

dims (approved):

PR #277, opened 2025-11-18
Use PATCH to update node labels instead of UPDATE

rajathagasthya:
PATCH is resilient to resource version conflict errors from API server.

### DIFF
--- a/pkg/mig/reconfigure/reconfigure.go
+++ b/pkg/mig/reconfigure/reconfigure.go
@@ -18,6 +18,7 @@ package reconfigure
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


pkg/mig/reconfigure/reconfigure.go
```diff
@@ -573,20 +575,16 @@ func (r *Reconfigure) setNodeLabel(key, value string) error {
...
-		node.Labels = make(map[string]string)
-	}
-
-	for key, value := range labels {
-		node.Labels[key] = value
+		return fmt.Errorf("failed to marshal patch data: %w", err)
 	}
 
-	_, err = r.clientset.CoreV1().Nodes().Update(r.ctx, node, metav1.UpdateOptions{})
+	_, err = r.clientset.CoreV1().Nodes().Patch(r.ctx, r.opts.NodeName, types.StrategicMergePatchType, patchData, metav1.PatchOptions{})
```
tariq1890: Let's use MergePatch instead of StrategicMergePatch. The motivation is to align with the `kubectl label` [behaviour](https://github.com/kubernetes/kubernetes/blob/5300466a5c8988b479a151ceb77f49dd00065c83/staging/src/k8s.io/kubectl/pkg/cmd/label/label.go#L351) 
```suggestion
	_, err = r.clientset.CoreV1().Nodes().Patch(r.ctx, r.opts.NodeName, types.MergePatchType, patchData, metav1.PatchOptions{})
```

rajathagasthya: Done.

PR #301, opened 2026-01-26
Unload nvidia_drm module for changing Mig mode

JunAr7112:


cmd/nvidia-mig-parted/util/device.go
```diff
@@ -145,12 +145,19 @@ func nvmlResetAllGPUs() (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("error getting GPU pci bus IDs: %v", err)
 	}
+
```
rajathagasthya: Unnecessary newline.

cmd/nvidia-mig-parted/util/device.go
```diff
@@ -145,12 +145,19 @@ func nvmlResetAllGPUs() (string, error) {
...
 	}
+	
 
 	if len(pciBusIDs) == 0 {
 		return "No GPUs to reset...", nil
 	}
 
-	cmd := exec.Command("nvidia-smi", "-r", "-i", strings.Join(pciBusIDs, ",")) //nolint:gosec
+	// Unload nvidia_drm module and its dependencies before reset to release GPU references
+	modprobeCmd := exec.Command("sudo", "modprobe", "-r", "nvidia_drm")
```
rajathagasthya: Does this assume the user has `sudo` access?

cmd/nvidia-mig-parted/util/device.go
JunAr7112: Removed

cmd/nvidia-mig-parted/util/device.go
JunAr7112: Yes. I will remove this. I don't think we need to handle root permissions here

cmd/nvidia-mig-parted/util/device.go
```diff
@@ -150,6 +150,12 @@ func nvmlResetAllGPUs() (string, error) {
 		return "No GPUs to reset...", nil
 	}
 
+	// Unload nvidia_drm module and its dependencies before reset to release GPU references
+	modprobeCmd := exec.Command("modprobe", "-r", "nvidia_drm") //nolint:gosec
```
cdesiniotis, line 154: I don't believe this is the right place (or only place) for this. Note, `nvmlResetAllGPUs()` is only called in`ResetAllGPUs()` which in turn is only called in `ApplyMigMode()`. Based on the bug description, we need to unload this module across all MIG reconfigurations, not just a mode change.

cdesiniotis: The current implementation addresses the case where a GPU reset is performed (e.g. when changing the MIG mode). But what about when the current MIG mode is correct, but we are attempting to apply a new MIG configuration? Does the `mig-parted apply` still work when the nvidia_drm module is loaded, or are we required to unload it first?

PR #303, opened 2026-01-27
Ensure maximum number of SM are created by mig manager

JunAr7112:


rajathagasthya:
@JunAr7112 Are we missing `go.mod` changes in the commit? Only the vendor file is updated.

rajathagasthya:
> I'm moving these changes to the nvlib library https://github.com/NVIDIA/go-nvlib/pull/85

You will need to update `go.mod` here that version once those changes are merged.

PR #312, opened 2026-02-10
Fallback to static MIG config when profiles are not discovered

rajathagasthya:
Older drivers such as R535 does not support querying MIG profiles when MIG mode is disabled. This change restores the static config that was previously removed, so MIG manager can fallback to that config when generation is not possible.

### DIFF
--- a/cmd/nvidia-mig-manager/main.go
+++ b/cmd/nvidia-mig-manager/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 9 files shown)


cmd/nvidia-mig-manager/main.go
```diff
@@ -375,6 +378,18 @@ func setupMigConfig(ctx context.Context, clientset *kubernetes.Clientset) (strin
...
 	if err != nil {
+		if errors.Is(err, discovery.ErrNoProfilesDiscovered) {
+			log.Warn("No MIG profiles discovered. Checking for default config")
+			defaultPath := os.Getenv("DEFAULT_CONFIG_FILE")
+			if defaultPath != "" {
+				if _, statErr := os.Stat(defaultPath); statErr == nil {
+					log.Infof("Using default config: %s", defaultPath)
+					return defaultPath, nil
+				}
+				log.Warnf("Default config file not found: %s", defaultPath)
```
tariq1890: Can we log `statErr` too here?

rajathagasthya: Added.

PR #332, opened 2026-03-13
Reduce redundant NVML Init/Shutdown cycles

rajathagasthya:
nvidia-mig-manager.service consumes excessive CPU time (~42s on DGX B300
with 8x B200 GPUs) for operations completing in ~9s wall time. The root
cause is 68-100 redundant NVML Init/Shutdown cycles per service run.

Each nvml.Init() triggers dlopen("libnvidia-ml.so.1") + 24 dlsym() calls
to resolve versioned API symbols — expensive on multi-GPU systems. The
overhead comes from two compounding patterns:

1. Every method on nvmlMigModeManager and nvmlMigConfigManager
   independently calls Init()/Shutdown(), despite callers already
   maintaining an initialized NVML instance at the command level.

2. Callers create new nvml.New() instances inside per-GPU loops,
   each triggering a full Init/Shutdown cycle including version checks.

Fix by:
- Accepting nvml.Interface in constructors (aligning real constructors
  with mock constructors that already accept it)
- Removing per-method Init/Shutdown from all 7 manager methods
- Removing manager creation out of per-device loops to create once
  per command

This reduces NVML Init/Shutdown from ~100 to 1 per command, cutting CPU
time by 4.7x and dlsym calls from 572 to 45 (12.7x reduction).

### DIFF
--- a/hack/benchmark-perf.sh
+++ b/hack/benchmark-perf.sh
@@ -1,0 +1,414 @@
+#!/usr/bin/env bash
+
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# benchmark-perf.sh — Benchmark nvidia-mig-parted performance.
+#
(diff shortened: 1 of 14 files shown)


hack/benchmark-perf.sh
```diff
@@ -0,0 +1,414 @@
+#!/usr/bin/env bash
```
rajathagasthya, line 1: This file was added to show how benchmarking was done. This will be removed.

hack/benchmark-compare.sh
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env bash
```
rajathagasthya: Same. This file will be removed too.

tariq1890 (approved):
Thanks for the great work on this @rajathagasthya !

hack/benchmark-compare.sh
rajathagasthya: Removed.

hack/benchmark-perf.sh
rajathagasthya: Keeping this one. This could be useful in the future.

Issue #329, opened 2026-03-11
[Bug]: nvidia-mig-manager in v0.13.1 is failing with "Error: failed to validate MIG configuration: exit status 1"


realjump:
**Describe the bug**
We have upgraded to the latest nvidia-gpu-operator version 25.10.1. In doing so, the image version of nvidia-mig-manager is set to nvcr.io/nvidia/cloud-native/k8s-mig-manager:v0.13.1.

While generating of the specification file via cdi generate via the system service it always failed after a kernel update and a corresponding reboot, we set about finding the cause.

Ultimately, we discovered that nvidia-mig-manager > v.12.x fails at startup or when mig-config labels are changed:

> nvidia-mig-manager IS_HOST_DRIVER=true
nvidia-mig-manager NVIDIA_DRIVER_ROOT=/
nvidia-mig-manager DRIVER_ROOT_CTR_PATH=/host
nvidia-mig-manager NVIDIA_DEV_ROOT=/
nvidia-mig-manager DEV_ROOT_CTR_PATH=/host
nvidia-mig-manager WITH_SHUTDOWN_HOST_GPU_CLIENTS=true
nvidia-mig-manager Starting nvidia-mig-manager
nvidia-mig-manager W0311 11:57:32.053406 1658603 client_config.go:667] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
nvidia-mig-manager time="2026-03-11T11:57:32Z" level=info msg="Waiting for change to 'nvidia.com/mig.config' label"
nvidia-mig-manager time="2026-03-11T11:57:32Z" level=info msg="Updating to MIG config: config13"
nvidia-mig-manager time="2026-03-11T11:57:32Z" level=info msg="Current value of 'nvidia.com/gpu.deploy.device-plugin=true'\n"
nvidia-mig-manager time="2026-03-11T11:57:32Z" level=info msg="Current value of 'nvidia.com/gpu.deploy.gpu-feature-discovery=true'\n"
nvidia-mig-manager time="2026-03-11T11:57:32Z" level=info msg="Current value of 'nvidia.com/gpu.deploy.dcgm-exporter=true'\n"
nvidia-mig-manager time="2026-03-11T11:57:32Z" level=info msg="Current value of 'nvidia.com/gpu.deploy.dcgm=true'\n"
[...]

PR #341, opened 2026-04-06
chroot host migPartedBinary compatibility fix

shivakunv:
refer: [329](https://github.com/NVIDIA/mig-parted/issues/329)

To Reproduce:

**Method 1:** 
 Install GPU Operator:

```
helm install --wait --generate-name \
  -n gpu-operator --create-namespace \
  nvidia/gpu-operator \
  --version v26.3.0 \
  --set mig.strategy=single
```

Exec into the MIG manager pod:

`kubectl exec -it -n gpu-operator mig-manager-.... -- sh`
Copy the binary to the host mount and run it through chroot:

```
mkdir -p /host/usr/local/nvidia/mig-manager
cp /usr/bin/nvidia-mig-parted /host/usr/local/nvidia/mig-manager/
chroot /host /usr/local/nvidia/mig-manager/nvidia-mig-parted assert -f /usr/local/nvidia/mig-manager/config.yaml -c config13

```
Observe:
```
/usr/local/nvidia/mig-manager/nvidia-mig-parted: /lib64/libc.so.6: version `GLIBC_2.32' not found
/usr/local/nvidia/mig-manager/nvidia-mig-parted: /lib64/libc.so.6: version `GLIBC_2.34' not found
```

**Method: 2:** 
Reproduce through the host-driver path:

Reproduce by simulating the host chroot path manually:

Install the NVIDIA driver on the RHEL 8 host:

```
sudo dnf -y install dkms kernel-devel-$(uname -r) kernel-headers-$(uname -r) gcc make elfutils-libelf-devel
sudo dnf -y module reset nvidia-driver
sudo dnf -y module enable nvidia-driver:580-dkms
sudo dnf -y install cuda-drivers
```

Install GPU Operator:

```
helm install --wait --generate-name \
  -n gpu-operator --create-namespace \
  nvidia/gpu-operator \
  --version v26.3.0 \
  --set driver.enabled=false \
  --set mig.strategy=single
```

Check that the host-driver path is enabled: `IS_HOST_DRIVER = true ==> WITH_SHUTDOWN_HOST_GPU_CLIENTS = true`  in logs or env :

Check the MIG manager logs:

`kubectl logs -n gpu-operator nvidia-mig-manager...`

Observe:

```
/usr/local/nvidia/mig-manager/nvidia-mig-parted: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /usr/local/nvidia/mig-manager/nvidia-mig-parted)
/usr/local/nvidia/mig-manager/nvidia-mig-parted: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /usr/local/nvidia/mig-manager/nvidia-mig-parted)
```

In your case, nvidia-driver is installed on the host, `IS_HOST_DRIVER=true,` which makes `WITH_SHUTDOWN_HOST_GPU_CLIENTS=true`. 
This causes nvidia-mig-manager to execute `nvidia-mig-parted `through the host path.
The failure is caused by a host userspace compatibility issue on `RHEL-8`: the newer `nvidia-mig-parted` binary requires a newer `glibc` than the host provides. 

working on a fix so that on older host systems nvidia-mig-manager does not execute nvidia-mig-parted through chroot.

### DIFF
--- a/cmd/nvidia-mig-manager/main.go
+++ b/cmd/nvidia-mig-manager/main.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -540,8 +541,14 @@ func migReconfigure(ctx context.Context, migConfigValue string, clientset *kuber
 		if err != nil {
 			return fmt.Errorf("failed to copy nvidia-mig-parted to host: %w", err)
 		}
-		migPartedBinary = strings.Split(hostMigPartedBinary, " ")
-		opts.MigConfigFile = filepath.Join(hostNvidiaDirFlag, "mig-manager", "config.yaml")
+
+		hostMigPartedBinaryPath := filepath.Join(hostNvidiaDirFlag, "mig-manager", "nvidia-mig-parted")
+		if hostSupportsMigParted(hostRootMountFlag, hostMigPartedBinaryPath) {
+			migPartedBinary = strings.Split(hostMigPartedBinary, " ")
+			opts.MigConfigFile = filepath.Join(hostNvidiaDirFlag, "mig-manager", "config.yaml")
+		} else {
+			log.Warn("Host userspace is incompatible with the bundled nvidia-mig-parted binary, falling back to in-container execution")
+		}
 	}
 
 	rcfg, err := reconfigure.New(ctx, clientset, migPartedBinary, opts)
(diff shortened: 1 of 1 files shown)


cmd/nvidia-mig-manager/main.go
```diff
@@ -540,8 +541,14 @@ func migReconfigure(ctx context.Context, migConfigValue string, clientset *kuber
...
 		}
-		migPartedBinary = strings.Split(hostMigPartedBinary, " ")
-		opts.MigConfigFile = filepath.Join(hostNvidiaDirFlag, "mig-manager", "config.yaml")
+
+		hostMigPartedBinaryPath := filepath.Join(hostNvidiaDirFlag, "mig-manager", "nvidia-mig-parted")
+		if hostSupportsMigParted(hostRootMountFlag, hostMigPartedBinaryPath) {
+			migPartedBinary = strings.Split(hostMigPartedBinary, " ")
+			opts.MigConfigFile = filepath.Join(hostNvidiaDirFlag, "mig-manager", "config.yaml")
+		} else {
+			log.Warn("Host userspace is incompatible with the bundled nvidia-mig-parted binary, falling back to in-container execution")
```
cdesiniotis, line 550: I don't think this is the correct way to go about this. When drivers are pre-installed, we do want to execute mig-parted on the host as we may need to manage the lifecycle of systemd services. I would prefer if we updated our build system so that our mig-parted binaries get built with older versions of GLIBC. @rajathagasthya can you look into this?

rajathagasthya: I agree we should build mig-parted with older versions of glibc instead of doing this. I can look into this.

cdesiniotis:
Closing as not planned.

Issue #157, opened 2024-12-31
mig config failed


MLintin:
i have two A30 gpu
if i config with below, it will success
```
version: v1
mig-configs:
  all-disabled:
  - devices: all
    mig-enabled: false
  node1:
  - devices:
    - 0
    mig-enabled: true
    mig-devices:
      1g.6gb: 4
``` 
i config with below, it will success too
```
version: v1
mig-configs:
  all-disabled:
  - devices: all
    mig-enabled: false
  node1:
  - devices:
    - 0
    - 1
    mig-enabled: true
    mig-devices:
      1g.6gb: 4
``` 

but if i config with below, it will failed
```
version: v1
mig-configs:
  all-disabled:
  - devices: all
    mig-enabled: false
  node1:
  - devices:
    - 1
    mig-enabled: true
    mig-devices:
      1g.6gb: 4
``` 

logs:
Applying the MIG mode change from the selected config to the node (and double checking it took effect)
If the -r option was passed, the node will be automatically rebooted if this is not successful
time="2024-12-31T06:44:06Z" level=debug msg="Parsing config file..."
time="2024-12-31T06:44:06Z" level=debug msg="Selecting specific MIG config..."
time="2024-12-31T06:44:06Z" level=debug msg="Running apply-start hook"
time="2024-12-31T06:44:06Z" level=debug msg="Checking current MIG mode..."
time="2024-12-31T06:44:08Z" level=debug msg="Walking MigConfig for (devices=[1])"
time="2024-12-31T06:44:08Z" level=debug msg="  GPU 1: 0x20B710DE"
time="2024-12-31T06:44:08Z" level=debug msg="    Asserting MIG mode: Enabled"
time="2024-12-31T06:44:08Z" level=debug msg="    MIG capable: true\n"
time="2024-12-31T06:44:08Z" level=debug msg="    Current MIG mode: Disabled"
time="2024-12-31T06:44:10Z" level=debug msg="Running pre-apply-mode hook"
time="2024-12-31T06:44:10Z" level=debug msg="Applying MIG mode change..."
time="2024-12-31T06:44:13Z" level=debug msg="Walking MigConfig for (devices=[1])"
time="2024-12-31T06:44:13Z" level=debug msg="  GPU 1: 0x20B710DE"
time="2024-12-31T06:44:13Z" level=debug msg="    MIG capable: true\n"
time="2024-12-31T06:44:13Z" level=debug msg="    Current MIG mode: Disabled"
[...]

PR #372, opened 2026-05-11
Update GetMigConfig to use GIProfileID from target device

JunAr7112:
This MR is in response to this issue: https://github.com/NVIDIA/mig-parted/issues/157?reload=1?reload=1. Essentially the issue we are seeing is that SetMigConfig(gpu, config) resolves MIG profile strings (e.g. "1g.6gb") to GIProfileID values globally across all GPUs, not from the target device. As such, if the GIProfileID of a partition in GPU 0 differs from the GIProfileID of a partition in GPU 1, the mig manager will error: Error getting GPU instance profile info for '1g.6gb': ERROR_NOT_SUPPORTED"

To fix this, a helper function resolveMigProfileOnDevice was created that will map a logical MIG profile (from config / Flatten) to the NVML GI/CI profile IDs for the given GPU. It will loop through the profiles in GetMigProfiles() and identify the profile that matches the mp.String() we are trying to create. We will then update with the correct profile info.

In order to test this change, func TestSetMigConfigResolvesMigProfileOnTargetGPU was defined. The function will create two dummy GPUs, and override the GI Profile ID for GPU 1.

local-agadiyar@ipp2-0139:~/mig-parted$ go test ./pkg/mig/config -run TestSetMigConfigResolvesMigProfileOnTargetGPU -count=1
ok      github.com/NVIDIA/mig-parted/pkg/mig/config     0.008s

### DIFF
--- a/pkg/mig/config/config.go
+++ b/pkg/mig/config/config.go
@@ -22,6 +22,7 @@ import (
 
 	log "github.com/sirupsen/logrus"
 
+	nvdevlib "github.com/NVIDIA/go-nvlib/pkg/nvlib/device"
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
 
 	"github.com/NVIDIA/mig-parted/internal/nvlib"
@@ -40,6 +41,20 @@ type nvmlMigConfigManager struct {
 }
 
 var _ Manager = (*nvmlMigConfigManager)(nil)
+
+// resolveMigProfileOnDevice maps a logical MIG profile (from config / Flatten) to the
+// NVML GI/CI profile IDs for the given GPU. Global ParseMigProfile can pick IDs from
+// another GPU when names collide across devices.
+func resolveMigProfileOnDevice(profiles []nvdevlib.MigProfile, mp *types.MigProfile) (*types.MigProfile, error) {
+	key := mp.String()
+	for _, p := range profiles {
+		if p.Matches(key) {
+			info := p.GetInfo()
+			return &types.MigProfile{MigProfileInfo: info}, nil
+		}
+	}
+	return nil, fmt.Errorf("MIG profile %q not found on this GPU", key)
+}
 
 func NewNvmlMigConfigManager(nvml nvml.Interface) Manager {
 	return &nvmlMigConfigManager{nvml, nvlib.NewMock(nvml)}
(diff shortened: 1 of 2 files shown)


pkg/mig/config/config.go
```diff
@@ -41,6 +42,29 @@ type nvmlMigConfigManager struct {
...
+// resolveMigProfileOnDevice maps a logical MIG profile (from config / Flatten) to the
+// NVML GI/CI profile IDs for the given GPU. Global ParseMigProfile can pick IDs from
+// another GPU when names collide across devices.
+func resolveMigProfileOnDevice(nvmllib nvml.Interface, device nvml.Device, mp *types.MigProfile) (*types.MigProfile, error) {
+	devLib := nvdevlib.New(nvmllib)
+	nvdevDevice, err := devLib.NewDevice(device)
+	if err != nil {
+		return nil, fmt.Errorf("error wrapping device for MIG profile resolution: %w", err)
+	}
+	profiles, err := nvdevDevice.GetMigProfiles()
```
rajathagasthya: This will construct `nvdev` wrappers every single time inside `iteratePermutationsUntilSuccess`, which defeats the purpose of a cache maintained in `nvdevDevice.GetMigProfiles()`. Can we resolve all the MIG profiles once before, keep them in a map, and then refer it inside `iteratePermutationsUntilSuccess`?

Something like:

```go
  resolved, err := resolveConfigProfilesOnDevice(m.nvml, device, config)
  if err != nil {
        return fmt.Errorf("error resolving MIG profiles on GPU %d: %w", gpu, err)
  }

  err = iteratePermutationsUntilSuccess(config, func(mps []*types.MigProfile) error {
        // ... existing clear loop unchanged ...

        for _, mp := range mps {
                r := resolved[mp.String()]

                giProfileInfo, ret := device.GetGpuInstanceProfileInfo(r.GIProfileID)
                // ... rest of inner loop uses `r`
        }
  })
```

JunAr7112: Sure I can switch to this format

JunAr7112: Moved profiles, err := nvdevDevice.GetMigProfiles() out of the helper function.

pkg/mig/config/config_test.go
```diff
@@ -91,6 +91,56 @@ func TestGetSetMigConfig(t *testing.T) {
...
 
+func TestSetMigConfigResolvesMigProfileOnTargetGPU(t *testing.T) {
+	types.SetMockNVdevlib()
+
+	server := dgxa100.New()
+	manager := &nvmlMigConfigManager{
+		nvml:  server,
+		nvlib: nvlib.NewMock(server),
+	}
+
```
rajathagasthya, line 112: Good test setup! But this is somewhat hard to understand without comments. Can you add comments to each setup step? For example, the step below sets `GPU_INSTANCE_PROFILE_1_SLICE` for GPU 1 to simulate a heterogeneous setup.

rajathagasthya (commented):
Overall I'm +1 on this localized fix. We can revisit the underlying go-nvml method(s) in https://github.com/NVIDIA/go-nvlib/issues/91 later since it has a larger blast radius, but this fix works.

Let's get another set of eyes before merging.

pkg/mig/config/config_test.go
JunAr7112: Alright I'll add comments to help clarify

rajathagasthya (approved):
Let's get another approval before merging.

cdesiniotis (approved):

PR #382, opened 2026-05-19
Support Read-Only, systemd-less Systems

### LINKED ISSUE #356: [Bug]: k8s-mig-manager v0.14.0 observes nvidia.com/mig.config label but does not apply geometry on Talos
**Describe the bug**
When running standalone `k8s-mig-manager` on Talos Linux with preinstalled NVIDIA drivers/toolkit from Talos system extensions, `v0.14.0` observes changes to the `nvidia.com/mig.config` node label but does not perform a real MIG geometry transition.

The underlying `nvidia-mig-parted` command works manually in the same pod. Downgrading the manager image to `v0.12.3-ubuntu20.04` makes label-driven reconfiguration work.

**To Reproduce**
MIG mode was enabled on both A100s. The node initially had this applied profile:

```text
nvidia.com/mig.config=all-1g.5gb
nvidia.com/mig.config.state=success
```

With `v0.14.0`, changing the label:

```bash
kubectl label node gx06 nvidia.com/mig.config=all-balanced --overwrite
```

updated the node label, but the hardware geometry did not change.

`k8s-mig-manager` logs stopped after:

```text
Using custom config: /mig-parted-config/config.yaml
Waiting for change to 'nvidia.com/mig.config' label
Updating to MIG config: all-1g.5gb
```

No later logs such as applying the config or successfully updating appeared.

`nvidia-smi -L` still showed the old `all-1g.5gb` layout.

Manual assert for the requested config failed:

```bash
kubectl exec -n nvidia-mig-manager ds/nvidia-mig-manager -- nvidia-mig-parted -d assert -f /mig-parted-config/config.yaml -c all-balanced
```

Output included:

```text
Not all GPUs match the specified config
Assertion failure: selected configuration not currently applied
```

However, manual apply from inside the same pod worked:

```bash
kubectl exec -n nvidia-mig-manager ds/nvidia-mig-manager -- nvidia-mig-parted -d apply -f /mig-parted-config/config.yaml -c all-balanced
```

**Expected behavior**
`k8s-mig-manager v0.14.0` should apply the selected profile when `nvidia.com/mig.config` changes, then set:

```text
nvidia.com/mig.config.state=success
```

**Environment (please provide the following information):**
- OS: Talos Linux v1.12.6
- Kubernetes: v1.34.6
- Node: `gx06`
- GPUs: 2x NVIDIA A100-PCIE-40GB
- NVIDIA driver/toolkit: provided by Talos system extensions (580.126.20)
- Kernel Version: 6.18.18-talos
- GPU Operator: not installed
- RuntimeClass: `nvidia`
- MIG Manager image that fails: `nvcr.io/nvidia/cloud-native/k8s-mig-manager:v0.14.0`
- MIG Manager image that works: `nvcr.io/nvidia/cloud-native/k8s-mig-manager:v0.12.3-ubuntu20.04`

## Notes

This appears specific to the standalone `k8s-mig-manager v0.14.0` controller path, not to `nvidia-mig-parted` itself. The same container could run `nvidia-mig-parted apply` manually.

One visible difference is that newer `v0.14.0` uses the Go-native reconfigure path, while `v0.12.3` uses the older shell-based `reconfigure-mig.sh` path.

Arc676:
# Motivation

Informally, this PR adds (partial) support for Talos. Closes #356.

More formally, this PR adds support for ~~systems with read-only filesystems and~~ systems that do not run `systemd`.

# Description

The MIG manager assumes that it will be able to copy the `mig-parted` binary to the host and use `systemd` to restart host-side GPU services. Neither of these is true for Talos, which is an immutable OS that doesn't run `systemd`. Proper Talos support would introduce a dependency on the Talos API, but that is beyond the scope of this PR and likely falls beyond the scope of what this tool should support.

~~This PR adds support for systems like Talos by introducing two new flags (both of which are required for Talos):~~

~~1. Identifying the host as read-only: prevent the manager from attempting to copy data to the host~~
~~2. Flagging the absence of `systemd`: tell the manager to skip all `systemd` operations that would otherwise cause the program to hang, since there would be no response on DBus~~

~~This PR includes `nil`-checks for the `systemd` manager that were not present before. In the original code, these checks are effectively unnecessary because this member is always initialized and the entire program blocks on this initialization if `systemd` is not present.~~

## Updated Description

In the interest of clarity for other users, here's an updated description of what this PR actually changes after all the discussion and how it relates to Talos.

This PR replaces the unconditional initialization of the `systemd` manager with a lazy initialization that includes a timeout. In a separate update to the GPU operator Helm chart, the `WITH_SHUTDOWN_HOST_GPU_CLIENTS` flag is decoupled from the flag indicating the presence of drivers on the host. The MIG manager already skips `systemd` operations when `WITH_SHUTDOWN_HOST_GPU_CLIENTS=false`. Together, these two changes remove the hard dependency on `systemd`, allowing the application to run on devices without `systemd`.

The decoupling of the flag also means that the MIG manager no longer tries to copy the `mig-parted` binary to the host just because host drivers are present. Note that this means that `WITH_SHUTDOWN_HOST_GPU_CLIENTS` must also be set to `false` when running on a read-only host. No changes in this PR are directly related to support for read-only systems.

# Improvements

This is the simplest possible solution to the problem described in the linked issue. All the MIG- and GPU-related operations work fine[^1] on Talos. We simply need to skip over the parts that can't work on Talos. The obvious alternatives or improvements over this PR are:

1. Specifically catching the "read-only FS" error when attempting to copy the binary instead of requiring a flag to skip the operation entirely
2. Detecting the presence or absence of `systemd`, either by inspecting the running processes or by introducing a timeout on the DBus connection, and adjusting accordingly, instead of requiring a flag

# Caveats

~~This PR exists more for discussion than with the goal of being merged.~~ These changes were made based on a very cursory reading and superficial understanding of the MIG manager. There is likely a cleaner and more elegant way to achieve this. However, I'll submit the patch as a proof-of-concept: by disabling the host-copies and all `systemd` features, the MIG manager works properly on Talos. This is, at least for us, an important starting point.

[^1]: CUDA validation yields `ERROR: init 250 result=11s`. I haven't yet figured out what this means, but so far it hasn't impacted the use of the GPU. The GPU workloads still run fine, as does the CUDA validation pod.

### DIFF
--- a/internal/systemd/systemd.go
+++ b/internal/systemd/systemd.go
@@ -35,10 +35,22 @@ type Manager struct {
 }
 
 // NewManager creates a new Manager instance
-func NewManager(ctx context.Context) (*Manager, error) {
-	conn, err := dbus.NewSystemConnectionContext(ctx)
+func NewManager(ctx context.Context, timeout time.Duration) (*Manager, error) {
+	deadline, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	conn, err := dbus.NewSystemConnectionContext(deadline)
 	if err != nil {
-		return nil, fmt.Errorf("failed to connect to systemd D-Bus: %w", err)
+		err = fmt.Errorf("failed to connect to systemd D-Bus: %w", err)
+		if timeout := deadline.Err(); timeout == context.DeadlineExceeded {
+			err = errors.Join(err, timeout)
+		}
+		return nil, err
+	}
+
+	conn, err = dbus.NewSystemConnectionContext(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to recreate connection with no timeout")
 	}
 
 	return &Manager{
--- a/internal/systemd/systemd.go
+++ b/internal/systemd/systemd.go
@@ -34,9 +34,11 @@ type Manager struct {
 	conn *dbus.Conn
 }
 
+const SystemdTimeout = 2 * time.Second
+
 // NewManager creates a new Manager instance
-func NewManager(ctx context.Context, timeout time.Duration) (*Manager, error) {
-	deadline, cancel := context.WithTimeout(ctx, timeout)
+func NewManager(ctx context.Context) (*Manager, error) {
+	deadline, cancel := context.WithTimeout(ctx, SystemdTimeout)
 	defer cancel()
 
 	conn, err := dbus.NewSystemConnectionContext(deadline)
@@ -50,7 +52,7 @@ func NewManager(ctx context.Context, timeout time.Duration) (*Manager, error) {
 
 	conn, err = dbus.NewSystemConnectionContext(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("failed to recreate connection with no timeout")
+		return nil, fmt.Errorf("failed to recreate connection with no timeout: %w", err)
 	}
 
 	return &Manager{
--- a/internal/systemd/systemd.go
+++ b/internal/systemd/systemd.go
@@ -27,44 +27,85 @@ import (
 	"github.com/coreos/go-systemd/v22/dbus"
 )
 
+// systemdConnectTimeout bounds how long NewManager waits for the systemd D-Bus
+// connection (dial plus auth handshake) to be established. On hosts without a
+// running systemd/D-Bus daemon the system bus socket may still exist - for
+// example when /run/dbus/system_bus_socket is bind-mounted from the host - but
+// never answer, in which case the underlying auth handshake would otherwise
+// block forever. Bounding it turns that indefinite hang into a fast, actionable
+// error.
+const systemdConnectTimeout = 10 * time.Second
+
 // Manager handles systemd operations using the D-Bus API
 type Manager struct {
 	ctx context.Context
 
-	conn *dbus.Conn
+	conn   *dbus.Conn
+	cancel context.CancelFunc
 }
 
-const SystemdTimeout = 2 * time.Second
-
-// NewManager creates a new Manager instance
+// NewManager creates a new Manager instance connected to the systemd system
+// bus. It fails fast, rather than blocking indefinitely, when the D-Bus socket
+// exists but no systemd daemon is answering (as happens on systemd-less hosts
+// such as Talos Linux).
 func NewManager(ctx context.Context) (*Manager, error) {
-	deadline, cancel := context.WithTimeout(ctx, SystemdTimeout)
-	defer cancel()
+	return newManagerWithTimeout(ctx, systemdConnectTimeout)
+}
 
-	conn, err := dbus.NewSystemConnectionContext(deadline)
-	if err != nil {
-		err = fmt.Errorf("failed to connect to systemd D-Bus: %w", err)
-		if timeout := deadline.Err(); timeout == context.DeadlineExceeded {
-			err = errors.Join(err, timeout)
+// newManagerWithTimeout is the testable core of NewManager, with the connection
+// timeout injected so tests do not have to wait for the production default.
+func newManagerWithTimeout(ctx context.Context, timeout time.Duration) (*Manager, error) {
+	// Derive a cancelable context for the connection. go-systemd/godbus tie the
+	// lifetime of the connection to this context and, crucially, spawn a watcher
+	// goroutine that closes the underlying socket as soon as the context is done.
+	// We rely on that to unblock a stuck auth handshake: if the dial does not
+	// complete within the timeout we cancel the context, the socket is closed,
+	// and the blocked read returns an error instead of hanging forever.
+	connCtx, cancel := context.WithCancel(ctx)
+
+	type result struct {
+		conn *dbus.Conn
+		err  error
+	}
+	ch := make(chan result, 1)
+	go func() {
+		conn, err := dbus.NewSystemConnectionContext(connCtx)
+		ch <- result{conn: conn, err: err}
+	}()
+
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+
+	select {
+	case res := <-ch:
+		if res.err != nil {
+			cancel()
+			return nil, fmt.Errorf("failed to connect to systemd D-Bus: %w", res.err)
 		}
-		return nil, err
+		// Success: the connection (and connCtx) must outlive this call, so we do
+		// not cancel here. cancel is retained and invoked by Close().
+		return &Manager{
+			ctx:    ctx,
+			conn:   res.conn,
+			cancel: cancel,
+		}, nil
+	case <-timer.C:
+		// The dial is stuck: the socket exists but nothing is answering, as on a
+		// systemd-less host. Cancel the connection context so the godbus watcher
+		// closes the socket and unblocks the goroutine above, then report a clear
+		// error instead of hanging.
+		cancel()
+		return nil, fmt.Errorf("timed out after %s connecting to systemd D-Bus: the system bus socket exists but is not responding (is this a systemd-less host?)", timeout)
 	}
-
-	conn, err = dbus.NewSystemConnectionContext(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to recreate connection with no timeout: %w", err)
-	}
-
-	return &Manager{
-		ctx:  ctx,
-		conn: conn,
-	}, nil
 }
 
 // Close closes the D-Bus connection
 func (sm *Manager) Close() error {
 	if sm.conn != nil {
 		sm.conn.Close()
+	}
+	if sm.cancel != nil {
+		sm.cancel()
 	}
 	return nil
 }
--- a/internal/systemd/systemd.go
+++ b/internal/systemd/systemd.go
@@ -27,13 +27,8 @@ import (
 	"github.com/coreos/go-systemd/v22/dbus"
 )
 
-// systemdConnectTimeout bounds how long NewManager waits for the systemd D-Bus
-// connection (dial plus auth handshake) to be established. On hosts without a
-// running systemd/D-Bus daemon the system bus socket may still exist - for
-// example when /run/dbus/system_bus_socket is bind-mounted from the host - but
-// never answer, in which case the underlying auth handshake would otherwise
-// block forever. Bounding it turns that indefinite hang into a fast, actionable
-// error.
+// systemdConnectTimeout limits how long NewManager waits for the systemd D-Bus
+// connection (dial plus auth handshake) to be established
 const systemdConnectTimeout = 10 * time.Second
 
 // Manager handles systemd operations using the D-Bus API
@@ -44,23 +39,11 @@ type Manager struct {
 	cancel context.CancelFunc
 }
 
-// NewManager creates a new Manager instance connected to the systemd system
-// bus. It fails fast, rather than blocking indefinitely, when the D-Bus socket
-// exists but no systemd daemon is answering (as happens on systemd-less hosts
-// such as Talos Linux).
 func NewManager(ctx context.Context) (*Manager, error) {
 	return newManagerWithTimeout(ctx, systemdConnectTimeout)
 }
 
-// newManagerWithTimeout is the testable core of NewManager, with the connection
-// timeout injected so tests do not have to wait for the production default.
 func newManagerWithTimeout(ctx context.Context, timeout time.Duration) (*Manager, error) {
-	// Derive a cancelable context for the connection. go-systemd/godbus tie the
-	// lifetime of the connection to this context and, crucially, spawn a watcher
-	// goroutine that closes the underlying socket as soon as the context is done.
-	// We rely on that to unblock a stuck auth handshake: if the dial does not
-	// complete within the timeout we cancel the context, the socket is closed,
-	// and the blocked read returns an error instead of hanging forever.
 	connCtx, cancel := context.WithCancel(ctx)
 
 	type result struct {
@@ -82,18 +65,12 @@ func newManagerWithTimeout(ctx context.Context, timeout time.Duration) (*Manager
 			cancel()
 			return nil, fmt.Errorf("failed to connect to systemd D-Bus: %w", res.err)
 		}
-		// Success: the connection (and connCtx) must outlive this call, so we do
-		// not cancel here. cancel is retained and invoked by Close().
 		return &Manager{
 			ctx:    ctx,
 			conn:   res.conn,
 			cancel: cancel,
 		}, nil
 	case <-timer.C:
-		// The dial is stuck: the socket exists but nothing is answering, as on a
-		// systemd-less host. Cancel the connection context so the godbus watcher
-		// closes the socket and unblocks the goroutine above, then report a clear
-		// error instead of hanging.
 		cancel()
 		return nil, fmt.Errorf("timed out after %s connecting to systemd D-Bus: the system bus socket exists but is not responding (is this a systemd-less host?)", timeout)
 	}
--- a/internal/systemd/systemd.go
+++ b/internal/systemd/systemd.go
@@ -39,22 +39,30 @@ type Manager struct {
 	cancel context.CancelFunc
 }
 
-func NewManager(ctx context.Context) (*Manager, error) {
-	return newManagerWithTimeout(ctx, systemdConnectTimeout)
+type result struct {
+	conn *dbus.Conn
+	err  error
 }
 
-func newManagerWithTimeout(ctx context.Context, timeout time.Duration) (*Manager, error) {
-	connCtx, cancel := context.WithCancel(ctx)
+// NewManager creates a new Manager instance with a timeout on the systemd connection
+func NewManager(ctx context.Context) (*Manager, error) {
+	return newManagerWithTimeout(ctx, systemdConnectTimeout, nil)
+}
 
-	type result struct {
-		conn *dbus.Conn
-		err  error
-	}
+func defaultConnect(connCtx context.Context, ch chan result) {
+	conn, err := dbus.NewSystemConnectionContext(connCtx)
+	ch <- result{conn: conn, err: err}
+}
+
+func newManagerWithTimeout(ctx context.Context, timeout time.Duration, connect func(context.Context, chan result)) (*Manager, error) {
+	connCtx, cancel := context.WithCancel(ctx)
 	ch := make(chan result, 1)
-	go func() {
-		conn, err := dbus.NewSystemConnectionContext(connCtx)
-		ch <- result{conn: conn, err: err}
-	}()
+
+	if connect == nil {
+		connect = defaultConnect
+	}
+
+	go connect(connCtx, ch)
 
 	timer := time.NewTimer(timeout)
 	defer timer.Stop()
@@ -72,6 +80,12 @@ func newManagerWithTimeout(ctx context.Context, timeout time.Duration) (*Manager
 		}, nil
 	case <-timer.C:
 		cancel()
+		select {
+		case res := <-ch:
+			if res.conn != nil {
+				res.conn.Close()
+			}
+		}
 		return nil, fmt.Errorf("timed out after %s connecting to systemd D-Bus: the system bus socket exists but is not responding (is this a systemd-less host?)", timeout)
 	}
 }
--- a/internal/systemd/systemd.go
+++ b/internal/systemd/systemd.go
@@ -80,12 +80,14 @@ func newManagerWithTimeout(ctx context.Context, timeout time.Duration, connect f
 		}, nil
 	case <-timer.C:
 		cancel()
-		select {
-		case res := <-ch:
+
+		go func() {
+			res := <-ch
 			if res.conn != nil {
 				res.conn.Close()
 			}
-		}
+		}()
+
 		return nil, fmt.Errorf("timed out after %s connecting to systemd D-Bus: the system bus socket exists but is not responding (is this a systemd-less host?)", timeout)
 	}
 }
--- a/internal/systemd/systemd.go
+++ b/internal/systemd/systemd.go
@@ -65,7 +65,7 @@ func newManagerWithTimeout(ctx context.Context, timeout time.Duration, connect f
 		case ch <- res:
 			// Connected within time limit
 		case <-connCtx.Done():
-			// Drain connection if obtained after timeout
+			// Explicitly drain connection if obtained after timeout
 			if conn != nil {
 				conn.Close()
 			}
@@ -87,6 +87,7 @@ func newManagerWithTimeout(ctx context.Context, timeout time.Duration, connect f
 			cancel: cancel,
 		}, nil
 	case <-timer.C:
+		// Canceling the connection context both disconnects a late result and aborts an ongoing connection attempt
 		cancel()
 		return nil, fmt.Errorf("timed out after %s connecting to systemd D-Bus: the system bus socket exists but is not responding (is this a systemd-less host?)", timeout)
 	}
--- a/internal/systemd/systemd.go
+++ b/internal/systemd/systemd.go
@@ -49,20 +49,28 @@ func NewManager(ctx context.Context) (*Manager, error) {
 	return newManagerWithTimeout(ctx, systemdConnectTimeout, nil)
 }
 
-func defaultConnect(connCtx context.Context, ch chan result) {
-	conn, err := dbus.NewSystemConnectionContext(connCtx)
-	ch <- result{conn: conn, err: err}
-}
-
-func newManagerWithTimeout(ctx context.Context, timeout time.Duration, connect func(context.Context, chan result)) (*Manager, error) {
+func newManagerWithTimeout(ctx context.Context, timeout time.Duration, connect func(context.Context) (*dbus.Conn, error)) (*Manager, error) {
 	connCtx, cancel := context.WithCancel(ctx)
 	ch := make(chan result, 1)
 
 	if connect == nil {
-		connect = defaultConnect
+		connect = dbus.NewSystemConnectionContext
 	}
 
-	go connect(connCtx, ch)
+	go func() {
+		// Block until connected or canceled
+		conn, err := connect(connCtx)
+		res := result{conn: conn, err: err}
+		select {
+		case ch <- res:
+			// Connected within time limit
+		case <-connCtx.Done():
+			// Drain connection if obtained after timeout
+			if conn != nil {
+				conn.Close()
+			}
+		}
+	}()
 
 	timer := time.NewTimer(timeout)
 	defer timer.Stop()
@@ -80,14 +88,6 @@ func newManagerWithTimeout(ctx context.Context, timeout time.Duration, connect f
 		}, nil
 	case <-timer.C:
 		cancel()
-
-		go func() {
-			res := <-ch
-			if res.conn != nil {
-				res.conn.Close()
-			}
-		}()
-
 		return nil, fmt.Errorf("timed out after %s connecting to systemd D-Bus: the system bus socket exists but is not responding (is this a systemd-less host?)", timeout)
 	}
 }
--- a/internal/systemd/systemd_test.go
+++ b/internal/systemd/systemd_test.go
@@ -1,0 +1,137 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package systemd
+
+import (
+	"context"
+	"net"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+)
+
+// setBusAddress points the D-Bus client library at socketPath for the duration
+// of the test. godbus resolves the system bus address from different
+// environment variables depending on the platform (DBUS_SYSTEM_BUS_ADDRESS on
+// Linux, DBUS_LAUNCHD_SESSION_BUS_SOCKET on macOS), so we set both to keep the
+// test portable between the CI runners and local development machines.
+func setBusAddress(t *testing.T, socketPath string) {
+	t.Helper()
+	t.Setenv("DBUS_SYSTEM_BUS_ADDRESS", "unix:path="+socketPath)
+	t.Setenv("DBUS_LAUNCHD_SESSION_BUS_SOCKET", socketPath)
+}
+
+// shortSocketPath returns a socket path in a fresh temp dir kept short enough to
+// stay under the ~104 byte sun_path limit for unix domain sockets (t.TempDir()
+// embeds the test name and can overflow that limit on macOS).
+func shortSocketPath(t *testing.T) string {
+	t.Helper()
+	dir, err := os.MkdirTemp("", "migp")
+	if err != nil {
+		t.Fatalf("failed to create temp dir: %v", err)
+	}
+	t.Cleanup(func() { _ = os.RemoveAll(dir) })
+	return filepath.Join(dir, "s")
+}
+
+// TestNewManagerWithTimeoutUnresponsiveSocket reproduces the systemd-less-host
+// failure mode: a system bus socket that exists (something is listening) but
+// never answers the D-Bus auth handshake. Before the fix this blocked forever;
+// NewManager must instead fail fast once the connect timeout elapses.
+func TestNewManagerWithTimeoutUnresponsiveSocket(t *testing.T) {
+	socketPath := shortSocketPath(t)
+
+	listener, err := net.Listen("unix", socketPath)
+	if err != nil {
+		t.Fatalf("failed to listen on unix socket %q: %v", socketPath, err)
+	}
+	defer listener.Close()
+
+	// Accept connections but never write a reply, mimicking a bind-mounted host
+	// socket with no systemd/D-Bus daemon behind it.
+	go func() {
+		for {
+			conn, err := listener.Accept()
+			if err != nil {
+				return
+			}
+			// Hold the connection open without responding.
+			t.Cleanup(func() { _ = conn.Close() })
+		}
+	}()
+
+	setBusAddress(t, socketPath)
+
+	const timeout = 300 * time.Millisecond
+	start := time.Now()
+	mgr, err := newManagerWithTimeout(context.Background(), timeout)
+	elapsed := time.Since(start)
+
+	if err == nil {
+		_ = mgr.Close()
+		t.Fatal("expected an error connecting to an unresponsive D-Bus socket, got nil")
+	}
+	if mgr != nil {
+		t.Errorf("expected a nil manager on error, got %v", mgr)
+	}
+	if !strings.Contains(err.Error(), "timed out") {
+		t.Errorf("expected a timeout error, got: %v", err)
+	}
+	// The call must return promptly after the timeout, not hang. A generous
+	// upper bound still catches a regression back to the indefinite block.
+	if elapsed > 5*time.Second {
+		t.Errorf("connect took %s, expected it to fail fast near the %s timeout", elapsed, timeout)
+	}
+}
+
+// TestNewManagerWithTimeoutMissingSocket verifies that a missing socket fails
+// immediately via the dial error path, well before the connect timeout - i.e.
+// the timeout is a backstop, not the primary error path.
+func TestNewManagerWithTimeoutMissingSocket(t *testing.T) {
+	// A path that does not exist: the dial fails right away.
+	setBusAddress(t, filepath.Join(t.TempDir(), "does-not-exist"))
+
+	const timeout = 10 * time.Second
+	start := time.Now()
+	mgr, err := newManagerWithTimeout(context.Background(), timeout)
+	elapsed := time.Since(start)
+
+	if err == nil {
+		_ = mgr.Close()
+		t.Fatal("expected an error connecting to a missing D-Bus socket, got nil")
+	}
+	if mgr != nil {
+		t.Errorf("expected a nil manager on error, got %v", mgr)
+	}
+	if strings.Contains(err.Error(), "timed out") {
+		t.Errorf("expected the immediate dial error, not the timeout error: %v", err)
+	}
+	if elapsed > 2*time.Second {
+		t.Errorf("connect to a missing socket took %s, expected it to fail immediately", elapsed)
+	}
+}
+
+// TestManagerCloseNil ensures Close is safe on a zero-value Manager (no
+// connection, no cancel func), matching how cleanup paths may call it.
+func TestManagerCloseNil(t *testing.T) {
+	var mgr Manager
+	if err := mgr.Close(); err != nil {
+		t.Errorf("Close on a zero-value Manager returned an error: %v", err)
+	}
+}
--- a/internal/systemd/systemd_test.go
+++ b/internal/systemd/systemd_test.go
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) NVIDIA CORPORATION.  All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,20 +26,12 @@ import (
 	"time"
 )
 
-// setBusAddress points the D-Bus client library at socketPath for the duration
-// of the test. godbus resolves the system bus address from different
-// environment variables depending on the platform (DBUS_SYSTEM_BUS_ADDRESS on
-// Linux, DBUS_LAUNCHD_SESSION_BUS_SOCKET on macOS), so we set both to keep the
-// test portable between the CI runners and local development machines.
 func setBusAddress(t *testing.T, socketPath string) {
 	t.Helper()
 	t.Setenv("DBUS_SYSTEM_BUS_ADDRESS", "unix:path="+socketPath)
 	t.Setenv("DBUS_LAUNCHD_SESSION_BUS_SOCKET", socketPath)
 }
 
-// shortSocketPath returns a socket path in a fresh temp dir kept short enough to
-// stay under the ~104 byte sun_path limit for unix domain sockets (t.TempDir()
-// embeds the test name and can overflow that limit on macOS).
 func shortSocketPath(t *testing.T) string {
 	t.Helper()
 	dir, err := os.MkdirTemp("", "migp")
@@ -50,10 +42,6 @@ func shortSocketPath(t *testing.T) string {
 	return filepath.Join(dir, "s")
 }
 
-// TestNewManagerWithTimeoutUnresponsiveSocket reproduces the systemd-less-host
-// failure mode: a system bus socket that exists (something is listening) but
-// never answers the D-Bus auth handshake. Before the fix this blocked forever;
-// NewManager must instead fail fast once the connect timeout elapses.
 func TestNewManagerWithTimeoutUnresponsiveSocket(t *testing.T) {
 	socketPath := shortSocketPath(t)
 
@@ -100,11 +88,7 @@ func TestNewManagerWithTimeoutUnresponsiveSocket(t *testing.T) {
 	}
 }
 
-// TestNewManagerWithTimeoutMissingSocket verifies that a missing socket fails
-// immediately via the dial error path, well before the connect timeout - i.e.
-// the timeout is a backstop, not the primary error path.
 func TestNewManagerWithTimeoutMissingSocket(t *testing.T) {
-	// A path that does not exist: the dial fails right away.
 	setBusAddress(t, filepath.Join(t.TempDir(), "does-not-exist"))
 
 	const timeout = 10 * time.Second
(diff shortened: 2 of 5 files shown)


linkages:
Just want to add my feedback to this. I just tested this with the following setup:

Environment:

- OS: Talos Linux v1.13.0
- Kubernetes: v1.34.0
- GPUs: NVIDIA B300 NVL and RTX 6000 Pro Server
- NVIDIA driver/toolkit: provided by Talos system extensions (580.159.03)
- Kernel Version: 6.18.29-talos
- GPU Operator: Helm chart v26.3.1

I had to build a new k8s-mig-manager based on @Arc676 repo. I then pushed it to docker.io/linkages/k8s-mig-manager:v0.14.1.

Then when I deploy the gpu-operator, I set the values for the helm chart using this:

```yaml
driver:
  enabled: false

toolkit:
  enabled: false

hostPaths:
  driverInstallDir: /usr/local

mig:
  strategy: mixed

migManager:
  enabled: true
  repository: docker.io/linkages
  version: v0.14.1
  env:
    - name: READONLY_ROOTFS
      value: "true"
    - name: SYSTEMD_UNAVAILABLE
      value: "true"

operator:
  cleanupCRD: true
```

I then set the `nvidia.com/mig.config` label on all nodes to `all-balanced` and the mig-manager did the right thing in waiting for all the operator components to stop and then it adjusted the MiG settings and restarted everything back up. Shortly after the gpu-feature-discovery controller set the correct labels on the nodes.

This was tested on 2 different types of nodes in the same cluster:

2 x Lenovo nodes with 8 x NVIDIA RTX PRO 6000 Blackwell Server Edition GPUs
and
1 x DGX B300 with 8 x NVIDIA B300 SXM6 AC

Thank you @Arc676 for this patch. I hope it or a more elegant version of this gets pulled upstream. For now this solves my problem.

rajathagasthya:
Thanks @Arc676 for tackling this! 

My preference is to replace the flag-based approach in favor of something with no new public surface area. The reason being `WITH_SHUTDOWN_HOST_GPU_CLIENTS=false` already encodes the intent "don't touch the host". The current bug is that `reconfigure.New()` unconditionally connects to systemd even when `WITH_SHUTDOWN_HOST_GPU_CLIENTS=false`. 

We can document that on platforms without systemd `WITH_SHUTDOWN_HOST_GPU_CLIENTS=false` is required. To fully close the loop on getting this to work on Talos, I would also suggest creating an issue to decouple `WITH_SHUTDOWN_HOST_GPU_CLIENTS` from `IS_HOST_DRIVER` in the GPU Operator helm chart.

In the meantime, we should remove the unconditional `systemd.NewManager()` call from `reconfigure.New()`, and replace it with lazy initialization and a context timeout. This way, when systemd isn't available, the MIG manager fails fast with a clear error instead of hanging silently.

Arc676:
Thanks for the feedback! I agree with your point regarding the public surface; adding flags was the easiest approach but both properties can be inferred from the system's behavior. Instead of requiring the user to set these flags, `nvidia-mig-manager` can set them automatically.

> In the meantime, we should remove the unconditional `systemd.NewManager()` call from `reconfigure.New()`, and replace it with lazy initialization and a context timeout. This way, when systemd isn't available, the MIG manager fails fast with a clear error instead of hanging silently.

Support for systems without `systemd` requires the option to disable the connection to `systemd` entirely. I'm not sure I understand what exactly you have in mind here.

What would be the advantage of a lazy init for the `systemd` manager? I suppose the startup would be slightly faster by skipping a step if the desired MIG configuration is already applied, but I think that determining the availability of `systemd` straight away makes more sense. In strictly managed environments, startup occurs at known times when the GPU operator is updated. Deferring the `systemd` check to when it's needed would mean that the latency would be incurred when the user attempts to change the MIG configuration. The first repartitioning operation after startup would be slower than the others.

We could keep the unconditional `systemd.NewManager` but introduce the timeout as you mentioned; if the connection times out, then we would set a flag to indicate that `systemd` is unavailable. In particular, the MIG manager should _not_ fail in this case, but perhaps output a warning to ensure that the user is aware. Or did you mean that `WITH_SHUTDOWN_HOST_GPU_CLIENTS=false` should be equivalent to "`systemd` unavailable"?

Unless you want to separate the features, I'd implement all these changes in this PR such that Talos support is covered.

I've created a new issue to track the change to the Helm chart per your suggestion.

Arc676:
I've adapted the implementation based on the above comments:

- Assume the Helm chart will be adapted such that `WITH_SHUTDOWN_HOST_GPU_CLIENTS` correctly reflects the intent "do not touch the host"; I've removed the flag for readonly hosts
- Since the `Reconfigure` object is recreated each time, attempts to persist that fail due to a read-only FS issue a warning but don't return an `error`. Caching this finding would have to occur at the program's top-level; I suppose this could be added, but with the above point there are no attempts to write to the host at this level.
- The DBus connection uses the same `Context` throughout; I didn't find a way to set a timeout on just the initial connection. DBus closes the network connection when the context times out and issues a corresponding warning that is _not_ wrapped with `context.DeadlineExceeded`. It's not particularly elegant but as a workaround I changed the initialization function to try twice: once with a timeout, after which `systemd` is flagged as unavailable (at least for the current reconfiguration attempt), and a second time with the parent context. As before, we'd need to query `systemd` outside the `Reconfigure` object to be able to cache the result. I've left in a flag to change the timeout, primarily to avoid having a fixed constant in the code. However, this does mean that _every_ reconfiguration will have to wait for this timeout.

Arc676:
Any update here? Not sure what's blocking the automated checks; I don't really use GitHub actions.

rajathagasthya:
@Arc676 This is on my list. I'm going to take a closer look this week. Thanks for your patience!

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -105,9 +108,14 @@ func New(ctx context.Context, clientset *kubernetes.Clientset, migPartedBinary [
 		_ = os.Setenv("DBUS_SYSTEM_BUS_ADDRESS", hostSystemBusAddress)
 	}
 
-	systemdManager, err := systemd.NewManager(ctx)
+	systemdManager, err := systemd.NewManager(ctx, time.Duration(opts.SystemdTimeout)*time.Second)
```
rajathagasthya: To expand on my earlier suggestion of lazy init of systemd manager:

1. We could remove the unconditional init in this `New()` method, move it to a cached getter method which all the callers that use systemd (`shutdownHostGPUClients`, `hostStartSystemdServices` and `persistConfig`) invoke. Each reconfiguration calls `New()` anyway, so latency doesn't change. 
2. On Talos, `shutdownHostGPUClients` and `hostStartSystemdServices` are never invoked when `WithShutdownHostGPUClients = true`. `persistConfig` only runs if `HOST_MIG_MANAGER_STATE_FILE` already exists on the host (the default value is `/etc/systemd/system/nvidia-mig-manager.service.d/override.conf` which is not applicable to Talos).
3. Right now, you're treating connection timeout as "systemd not present" and ignoring it. On systems that have systemd, connection timeouts should be hard failures and shouldn't be equated to systemd not being present.

I think the code could look something like this:

```go
// Cached getter
func (r *Reconfigure) getSystemdManager() (*systemd.Manager, error) {
      if r.systemdManager != nil {
              return r.systemdManager, nil
      }
      m, err := systemd.NewManager(r.ctx) // connect timeout lives inside NewManager
      if err != nil {
              return nil, fmt.Errorf("failed to connect to systemd (if the host does not run systemd, set WITH_SHUTDOWN_HOST_GPU_CLIENTS=false): %w", err)
      }
      r.systemdManager = m
      return m, nil
}

// Caller
func (r *Reconfigure) shutdownHostGPUClients() error {
      m, err := r.getSystemdManager()
      if err != nil {
              return err
      }
      stoppedServices, err := m.StopSystemdServices(services)
      ...
}
```

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -322,9 +330,17 @@ Environment="MIG_PARTED_SELECTED_CONFIG=%s"
 
 	// Write config to file
 	if err := os.WriteFile(stateFilePath, []byte(config), 0600); err != nil {
+		if errors.Is(err, syscall.EROFS) {
```
rajathagasthya: Is this necessary? As mentioned above, `persistConfig` will never be called in non-systemd systems.

rajathagasthya (commented):
@Arc676 Thanks for your patience and for reworking this! I expanded on some of my earlier suggestions in this review. 

As for https://github.com/NVIDIA/gpu-operator/issues/2501, I think the decoupling can be done in a non-breaking way in the entrypoint script like below:
```bash
WITH_SHUTDOWN_HOST_GPU_CLIENTS=${WITH_SHUTDOWN_HOST_GPU_CLIENTS:-$IS_HOST_DRIVER}
```

pkg/mig/reconfigure/reconfigure.go
Arc676: I included the check to make the change more generic, since "read-only FS" and "no `systemd`" could be considered separate features. Talos just happens to have both. Just as the existing implementation didn't foresee the incompatibility with Talos, I felt that it would make more sense to avoid coupling these two properties just because Talos has both in case some future setup has one property but not the other.

However, I overlooked the fact that persistence is skipped if the file doesn't already exist; thanks for pointing this out. In that case, the check is incorrect; if the user specifies a path that exists but cannot be written to, then that should be an error. I'll remove this change.

lexfrei:
I opened #431 for the same bug before seeing this thread, and was pointed here — makes sense to land the fix in this PR rather than compete, since it's older and already hardware-tested. Happy to fold my work in (co-author, or just take the pieces), or step back if you'd rather carry it yourself.

I did end up at exactly the refactor requested in the 07-08 review, so it might save a round-trip. In #431:

- `reconfigure.New()` no longer dials D-Bus; the connection moved to a lazy cached getter (matches the sketch in the review), so on a systemd-less host with `WITH_SHUTDOWN_HOST_GPU_CLIENTS=false` it's never dialed and the reconfigure completes.
- The connect timeout lives inside `NewManager` and returns a hard error rather than nil, so on a real systemd host an unresponsive socket surfaces as an error instead of being silently treated as "no systemd". It's a single dial (avoiding the current double-dial, which leaks the first connection and then re-dials without a timeout), and the timeout is an internal constant — no new flag or public surface.
- Unit tests cover the timeout path, the missing-socket path, lazy construction, and nil-safe cleanup.

The diff is on #431 as a reference. Glad to push these here as commits with you as co-author, or hand them over — whatever's easiest. Just say which you prefer.

pkg/mig/reconfigure/reconfigure.go
Arc676: Regarding point 3: this is true. Using the timeout was the simplest way to check for `systemd` without introducing much additional logic. However, it occurs to me that the `systemd` dependency is actually tied to features behind the `WithShutdownHostGPUClients` flag. This means that `mig-parted` doesn't need a `systemd` detection mechanism if we instead rely on the flag.

I've implemented the lazy-init as you suggested and removed the soft-fail for timeouts. I've kept the timeout check itself since this prevents the program from hanging in the event of an actual DBus connection issue. The new implementation thus assumes that `WithShutdownHostGPUClients` will be set only when `systemd` is running. Otherwise, the reconfiguration will always fail due to connection timeout. With the flags set properly, none of the `systemd`-dependent calls can be reached with the flag is disabled, so the initializer is never called.

As a side note for anyone using this patch on their own setup: you have to patch the operator manually to use this feature branch now.

<details>
<summary>Patch Details</summary>

The new changes make this build of `mig-parted` dependent on the decoupling of the flags in the chart. Until the chart issue is resolved, `mig-parted` can only be used in this form if the entrypoint script is patched in the operator image (which can be done using a volume mount and a custom `ConfigMap`).

Custom entrypoint:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: nvidia-mig-manager-custom-entrypoint
data:
  configmap.yaml: |
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: nvidia-mig-manager-entrypoint
      namespace: "FILLED BY THE OPERATOR"
      labels:
        app: nvidia-mig-manager
    data:
      entrypoint.sh: |
        #!/bin/sh

        until [ -f /run/nvidia/validations/driver-ready ]
        do
          echo "waiting for the driver validations to be ready..."
          sleep 5
        done

        set -o allexport
        cat /run/nvidia/validations/driver-ready
        . /run/nvidia/validations/driver-ready
          
        # manually export additional envs required by mig-manager
        export WITH_SHUTDOWN_HOST_GPU_CLIENTS=${WITH_SHUTDOWN_HOST_GPU_CLIENTS:-$IS_HOST_DRIVER}
        echo "WITH_SHUTDOWN_HOST_GPU_CLIENTS=$WITH_SHUTDOWN_HOST_GPU_CLIENTS"

        echo "Starting nvidia-mig-manager"
        exec nvidia-mig-manager
```

Mounting into the operator (just showing the altered bits):

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: gpu-operator
spec:
  template:
    spec:
      volumes:
        - name: mig-entrypoint
          configMap:
            name: nvidia-mig-manager-custom-entrypoint
      containers:
        - name: gpu-operator
          volumeMounts:
            - name: mig-entrypoint
              mountPath: /opt/gpu-operator/state-mig-manager/0420_configmap.yaml
              subPath: configmap.yaml
```

</details>

Arc676 (commented):
@lexfrei thanks for reaching out! I actually prefer the timer-based approach (I hadn't thought of that); it's a lot cleaner than opening two connections. I also hadn't implemented any unit tests, since I tested the build in the cluster directly, so that's a useful addition for the pipeline here on GitHub. I cherry-picked your commits into my branch and updated the trailers in case the code-signing checks care about authorship. Conveniently, we used similar identifiers so there were almost no conflicts.

@rajathagasthya I went with the assumption that having unit tests is better than not having them. However, the commits are still distinct, so if you would prefer to review the tests separately, I can easily reset the branch to its `HEAD~2` and re-separate the two PRs. Let me know what you think makes the most sense. I think at this point, by line count, this PR mostly contains the content of the other PR. They do the same thing so I don't really mind which one you'd prefer to follow going forward.

lexfrei:
@Arc676 thanks for picking up my commits from #431 — the lazy-init shape is right.

One regression to flag, in `cleanup()`:

```go
func (r *Reconfigure) cleanup() {
	if mgr, _ := r.getSystemdManager(); mgr != nil {
		mgr.Close()
	}
}
```

`getSystemdManager()` only returns from cache when a connection already exists; otherwise it dials. So on a systemd-less host — where nothing ever established a manager — `cleanup()` now dials the D-Bus socket, waits out the timeout, gets an error, and closes nothing. That's exactly the path the lazy-init exists to remove, and it's back on precisely the hosts this PR targets. The original guard was load-bearing:

```go
if r.systemdManager != nil {
	r.systemdManager.Close()
}
```

Reading `r.systemdManager` directly keeps cleanup side-effect free. `TestNewDoesNotConnectSystemd` won't catch this — it only pins `New()` — so a test calling `cleanup()` on a nil manager would be worth adding alongside the fix.

Separately, on `DefaultSystemdTimeout = 1.0`: that's a lot tighter than the 10s in the original. A busy host can lose a 1s D-Bus connect race legitimately, and anything that previously succeeded slowly would now fail instead. It's overridable, but 1s as the default looks likely to produce spurious failures.

Arc676:
@lexfrei thanks for catching the `cleanup` regression. I've pushed the correction.

The default timeout of one second is actually outdated; I adopted your implementation of the manager setup, which does not propagate the user's command line options. Hence, the timeout is actually fixed at 10 seconds. I've removed the obsolete constant. I can put back the command line parameter if needed.

lexfrei:
Checked your cleanup correction — looks right, thanks. On the timeout: I'd keep it as a fixed internal constant rather than reintroducing the CLI parameter. A connect timeout is a liveness bound, not a tuning knob — 10s is comfortably above any healthy D-Bus handshake, and one less flag keeps the public surface smaller. If someone ever shows a real environment where 10s is wrong, that's the moment to add the knob.

rajathagasthya:
> thanks for catching the cleanup regression. I've pushed the correction.

@Arc676 I do not see this correction in code; there are no new commits since then. What am I missing?

Arc676:
@rajathagasthya oops, this is embarrassing. I pushed to a different remote and didn't update the GitHub fork. Now the branch is updated (and I've set the local upstream to track this branch).

rajathagasthya:
All good, thanks @Arc676! We are in a code freeze right now for a GPU Operator release, so I'll hold off an approval for now, but it looks good to me. 

Would you also like to make the corresponding GPU Operator chart change?

```sh
WITH_SHUTDOWN_HOST_GPU_CLIENTS=${WITH_SHUTDOWN_HOST_GPU_CLIENTS:-$IS_HOST_DRIVER}
```

internal/systemd/systemd_test.go
```diff
@@ -0,0 +1,137 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
tariq1890:
```suggestion
 * Copyright (c) NVIDIA CORPORATION.  All rights reserved.
```

internal/systemd/systemd.go
```diff
@@ -27,31 +27,86 @@ import (
...
+
+// newManagerWithTimeout is the testable core of NewManager, with the connection
+// timeout injected so tests do not have to wait for the production default.
+func newManagerWithTimeout(ctx context.Context, timeout time.Duration) (*Manager, error) {
+	// Derive a cancelable context for the connection. go-systemd/godbus tie the
+	// lifetime of the connection to this context and, crucially, spawn a watcher
+	// goroutine that closes the underlying socket as soon as the context is done.
+	// We rely on that to unblock a stuck auth handshake: if the dial does not
+	// complete within the timeout we cancel the context, the socket is closed,
+	// and the blocked read returns an error instead of hanging forever.
```
tariq1890: can we try to shorten this?

internal/systemd/systemd.go
```diff
@@ -27,31 +27,86 @@ import (
...
+// such as Talos Linux).
 func NewManager(ctx context.Context) (*Manager, error) {
-	conn, err := dbus.NewSystemConnectionContext(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to connect to systemd D-Bus: %w", err)
+	return newManagerWithTimeout(ctx, systemdConnectTimeout)
+}
+
+// newManagerWithTimeout is the testable core of NewManager, with the connection
+// timeout injected so tests do not have to wait for the production default.
```
tariq1890: These comments are not necessary IMO.

internal/systemd/systemd.go
```diff
@@ -27,31 +27,86 @@ import (
...
+		return &Manager{
+			ctx:    ctx,
+			conn:   res.conn,
+			cancel: cancel,
+		}, nil
+	case <-timer.C:
+		// The dial is stuck: the socket exists but nothing is answering, as on a
+		// systemd-less host. Cancel the connection context so the godbus watcher
+		// closes the socket and unblocks the goroutine above, then report a clear
+		// error instead of hanging.
```
tariq1890: Not required, let's remove

internal/systemd/systemd.go
```diff
@@ -27,31 +27,86 @@ import (
...
-		conn: conn,
-	}, nil
+	select {
+	case res := <-ch:
+		if res.err != nil {
+			cancel()
+			return nil, fmt.Errorf("failed to connect to systemd D-Bus: %w", res.err)
+		}
+		// Success: the connection (and connCtx) must outlive this call, so we do
+		// not cancel here. cancel is retained and invoked by Close().
```
tariq1890: not required, let's remove

internal/systemd/systemd_test.go
```diff
@@ -0,0 +1,137 @@
...
+	if strings.Contains(err.Error(), "timed out") {
+		t.Errorf("expected the immediate dial error, not the timeout error: %v", err)
+	}
+	if elapsed > 2*time.Second {
+		t.Errorf("connect to a missing socket took %s, expected it to fail immediately", elapsed)
+	}
+}
+
+// TestManagerCloseNil ensures Close is safe on a zero-value Manager (no
+// connection, no cancel func), matching how cleanup paths may call it.
```
tariq1890: unnecessary, let's remove

internal/systemd/systemd_test.go
```diff
@@ -0,0 +1,137 @@
...
+	// The call must return promptly after the timeout, not hang. A generous
+	// upper bound still catches a regression back to the indefinite block.
+	if elapsed > 5*time.Second {
+		t.Errorf("connect took %s, expected it to fail fast near the %s timeout", elapsed, timeout)
+	}
+}
+
+// TestNewManagerWithTimeoutMissingSocket verifies that a missing socket fails
+// immediately via the dial error path, well before the connect timeout - i.e.
+// the timeout is a backstop, not the primary error path.
```
tariq1890: unnecessary, let's remove

internal/systemd/systemd_test.go
```diff
@@ -0,0 +1,137 @@
...
+		t.Fatalf("failed to create temp dir: %v", err)
+	}
+	t.Cleanup(func() { _ = os.RemoveAll(dir) })
+	return filepath.Join(dir, "s")
+}
+
+// TestNewManagerWithTimeoutUnresponsiveSocket reproduces the systemd-less-host
+// failure mode: a system bus socket that exists (something is listening) but
+// never answers the D-Bus auth handshake. Before the fix this blocked forever;
+// NewManager must instead fail fast once the connect timeout elapses.
```
tariq1890: unnecessary, let's remove

internal/systemd/systemd_test.go
```diff
@@ -0,0 +1,137 @@
...
+// test portable between the CI runners and local development machines.
+func setBusAddress(t *testing.T, socketPath string) {
+	t.Helper()
+	t.Setenv("DBUS_SYSTEM_BUS_ADDRESS", "unix:path="+socketPath)
+	t.Setenv("DBUS_LAUNCHD_SESSION_BUS_SOCKET", socketPath)
+}
+
+// shortSocketPath returns a socket path in a fresh temp dir kept short enough to
+// stay under the ~104 byte sun_path limit for unix domain sockets (t.TempDir()
+// embeds the test name and can overflow that limit on macOS).
```
tariq1890: unnecessary, let's remove

internal/systemd/systemd_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"strings"
+	"testing"
+	"time"
+)
+
+// setBusAddress points the D-Bus client library at socketPath for the duration
+// of the test. godbus resolves the system bus address from different
+// environment variables depending on the platform (DBUS_SYSTEM_BUS_ADDRESS on
+// Linux, DBUS_LAUNCHD_SESSION_BUS_SOCKET on macOS), so we set both to keep the
+// test portable between the CI runners and local development machines.
```
tariq1890: please consider shortening this or just removing it

pkg/mig/reconfigure/reconfigure_test.go
```diff
@@ -17,9 +17,69 @@
...
+	"context"
+	"path/filepath"
 	"testing"
 )
 
+// TestNewDoesNotConnectSystemd pins the core of the systemd-less-host fix:
+// constructing a Reconfigure must not dial the host's systemd D-Bus. On hosts
+// where the system bus socket exists but nothing answers (e.g. Talos Linux),
+// an eager connection here would block forever, so the connection has to be
+// established lazily instead.
```
tariq1890: same comment as before

pkg/mig/reconfigure/reconfigure_test.go
```diff
@@ -17,9 +17,69 @@
...
+		t.Fatal("New returned a nil Reconfigure")
+	}
+	if r.systemdManager != nil {
+		t.Error("New must not establish a systemd D-Bus connection; systemdManager should be nil")
+	}
+}
+
+// TestSystemdMgrErrorIsNotCached verifies that when the lazy systemd connection
+// fails, the error is surfaced and nothing is cached, so a later call can try
+// again rather than reusing a broken (nil) manager.
```
tariq1890: same as before

pkg/mig/reconfigure/reconfigure_test.go
```diff
@@ -17,9 +17,69 @@
...
+		t.Errorf("expected a nil manager on error, got %v", mgr)
+	}
+	if r.systemdManager != nil {
+		t.Error("a failed connection must not be cached; systemdManager should remain nil")
+	}
+}
+
+// TestCleanupWithNilManager ensures the deferred cleanup is safe when a
+// reconfiguration completes without ever needing systemd (the common
+// systemd-less-host path, where the manager is never created).
```
tariq1890: same as before, let's remove these comments

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -98,25 +98,47 @@ type Reconfigure struct {
...
+}
+
+// getSystemdManager lazily establishes and caches the connection to the host's systemd
+// D-Bus. It is only ever called from code paths that genuinely need systemd:
+// persisting the host mig-manager state file (which triggers a daemon-reload)
+// and stopping/starting host GPU client services. On hosts where none of those
+// run - notably when WITH_SHUTDOWN_HOST_GPU_CLIENTS is false and no host state
+// file exists - the D-Bus socket is never dialed, so a systemd-less host does
+// not hang and the MIG geometry change (applied by the nvidia-mig-parted
+// binary, not via systemd) still completes.
```
tariq1890: consider shortening

tariq1890:
Great work on this @Arc676 !

The PR looks good to me for the most part. My only ask is that we reduce the number of comments. In general, let's use comments sparingly; we reserve comments to provide context or information when it's not obvious after reading the code.

tariq1890:
Can you rebase your PR? I think that will resolve the stuck `Check third party notices` check.

internal/systemd/systemd.go
Arc676: I removed the comment entirely, since `context.WithCancel` is self-documenting and the rest of the block is just an explanation of why we need it. The intent should be clear from the code and worst case `git blame` will lead users to this discussion.

internal/systemd/systemd.go
```diff
@@ -27,31 +27,63 @@ import (
...
 type Manager struct {
 	ctx context.Context
 
-	conn *dbus.Conn
+	conn   *dbus.Conn
+	cancel context.CancelFunc
 }
 
-// NewManager creates a new Manager instance
 func NewManager(ctx context.Context) (*Manager, error) {
```
rajathagasthya, line 42: Let's the restore the doc comment on this exported function.

```go
// NewManager creates a new Manager instance
```

rajathagasthya (commented):
@Arc676 Could you squash your commits into a single commit?

sylvesterkaczmarek (commented):
There is still a timeout race that can leak a live D-Bus connection. If the timer/context branch wins just as `NewSystemConnectionContext` succeeds, the goroutine can enqueue a non-nil `*dbus.Conn` into the buffered result channel after `NewManager` has returned; go-systemd requires the successful connection to be closed, and nothing drains or closes this late result. Please make result ownership explicit so a connection arriving after cancellation is closed, and add a race-oriented test.

Arc676:
@sylvesterkaczmarek that's a good catch. Although I'm not sure how one would implement a race-oriented test. I guess one option would be to copy the `newManagerWithTimeout` implementation into the test and artificially force the codepath where the race condition is met. Or did you have another idea?

sylvesterkaczmarek:
I would avoid copying `newManagerWithTimeout` into the test, because that can let the test diverge from the production ownership logic. The production fix can make the timeout branch explicitly drain the single late result after `cancel()` and close any non-nil connection it receives. For deterministic coverage, factor the connect call behind a small injected connector used only by an internal helper, e.g. `newManagerWithTimeoutWithConnector(ctx, timeout, connect)`. The test connector can block until the timeout has won and then release a successful result; the assertion is that the late result is consumed/closed rather than left in the buffered channel. That makes the race deterministic without timing sleeps and tests the real helper.

Arc676:
@sylvesterkaczmarek sounds good to me. I've added a corresponding test case. Let me know what you think.

@rajathagasthya I put back the comment on the `NewManager` function. Once we have a final version, I'll squash all the commits. Do you have a preference regarding commit metadata (specifically the date)? Squash-merges on GitLab take the merge-time as the final commit date, which I personally dislike, since this means delays in reviews end up affecting the commit metadata. Squashing everything into the first commit also erases the duration of any follow-up changes, since `git` just takes the first commit's date. Usually I squash locally and use the most meaningful date among commits (typically the largest non-fixup commit). But I'm happy to squash however you feel is most appropriate for the repository conventions.

sylvesterkaczmarek (commented):
Thanks for adding the race-oriented coverage. The late-result ownership issue is improved, but the timeout branch currently calls `cancel()` and then performs a blocking receive from `ch` before returning. If the connector does not promptly honor cancellation, `NewManager` can still block past the advertised timeout. I would keep the timeout return independent of connector completion, then drain and close any late non-nil connection asynchronously. A deterministic injected connector that waits until explicitly released after the timeout would pin both properties: the call returns on time, and the late connection is still closed.

Arc676:
The blocking receive is my mistake; I don't do a lot of async. There should have been a `default` case. Although that wouldn't catch anything arriving even later. Honestly I don't know how to implement what you're suggesting. I'd appreciate a code suggestion or inline comment.

I asked GPT and got

```go
case <-timer.C:
    cancel()

    go func() {
        res := <-ch

        if res.conn != nil {
            _ = res.conn.Close()
        }
    }()

    return nil, timeoutErr
```

Putting aside the invalid `_ =`, if the connector never respects the cancellation, this would just continuously spawn new goroutines until the manager eventually exited. Or is the assumption that cancellation will always eventually be respected? The unit test as written does eventually clean up the late connection, but by introducing further delays, it's possible to make the connect call take so long that the test exits before the cleanup code is reached. This would simulate the case where the connector doesn't respect cancellation.

In the interest of time, I've pushed the corrected version of GPT's suggestion.

sylvesterkaczmarek:
Yes. I would move ownership of the late connection into the worker rather than start a second goroutine that can wait on `ch` forever.

The cleanest shape is to make the injected connector return the result, then use an unbuffered result channel:

```go
type connectFunc func(context.Context) (*dbus.Conn, error)

ch := make(chan result)
connCtx, cancel := context.WithCancel(ctx)

go func() {
    conn, err := connect(connCtx)
    res := result{conn: conn, err: err}
    select {
    case ch <- res:
        // NewManager owns conn now.
    case <-connCtx.Done():
        if conn != nil {
            conn.Close()
        }
    }
}()
```

Then the timeout branch only calls `cancel()` and returns. If the connect completes later, the worker sees cancellation and closes the connection itself; there is no blocking receive and no extra drain goroutine to leak. The deterministic test connector can wait for `ctx.Done()`, then return a successful connection and assert it is eventually closed.

Arc676:
Further testing has revealed that the original assumption about leaking a `dbus` connection was not quite correct. The `dbus.Conn` actually disconnects itself when its context is canceled. This means that, even without the changes we've been discussing, a race between the timer and the connector will not lead to a dangling connection. I've added a new test to verify this behavior in case it ever changes.

In the spirit of your comment regarding explicit ownership, I've made the change you suggested regardless of this new information. While running this code locally, I've found that the race-oriented test is quite flaky, albeit not in its results. The test always passes because of the aforementioned behavior but the exact codepath is not consistent. Adding debug prints in various locations is enough to alter the timing such that the program follows a different codepath. In particular, the `select` blocks in the goroutine and the main body of `newManagerWithTimeout` do not show the same behavior with respect to the timeout: it is possible for the worker goroutine to follow the `ch <- res` case even though the function body follows the case where the timer is triggered. The `dbus` connection is still closed when `cancel` is called, so skipping the drain has no impact.

I couldn't find an email of yours for the coauthor trailer so I picked a random GitHub-assigned one from one of your repositories. If you'd prefer that I include different metadata (or not attribute coauthorship at all), let me know and I'll update it when I squash the commits at the end of the review process.

sylvesterkaczmarek (commented):
Rechecked the current head. The timeout path now cancels the production D-Bus connection context, the worker owns any late result, and the added test confirms cancellation disconnects the connection. My earlier late-connection leak concern is resolved.

internal/systemd/systemd_test.go
```diff
@@ -0,0 +1,171 @@
...
+func TestManagerRace(t *testing.T) {
+	const timeout = 1 * time.Second
+
+	var conn *dbus.Conn
+
+	delayedConnect := func(ctx context.Context) (*dbus.Conn, error) {
+		var err error
+
+		// Intercept the connection so we can check its state afterwards
+		conn, err = dbus.NewSystemConnectionContext(ctx)
```
rajathagasthya, line 141: Does this initiate a real systemd connection? If so, the tests will fail on a machine with no systemd. Should this be pointed at a fake socket?

internal/systemd/systemd_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	if conn.Connected() {
+		t.Errorf("expected connection to be dropped after timeout")
+	}
+}
+
+func TestCancelDbus(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	conn, err := dbus.NewSystemConnectionContext(ctx)
```
rajathagasthya, line 165: Same here.

internal/systemd/systemd.go
```diff
@@ -27,31 +27,80 @@ import (
...
-	conn, err := dbus.NewSystemConnectionContext(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to connect to systemd D-Bus: %w", err)
+	return newManagerWithTimeout(ctx, systemdConnectTimeout, nil)
+}
+
+func newManagerWithTimeout(ctx context.Context, timeout time.Duration, connect func(context.Context) (*dbus.Conn, error)) (*Manager, error) {
+	connCtx, cancel := context.WithCancel(ctx)
+	ch := make(chan result, 1)
```
rajathagasthya: Should this still be a buffered channel? Correct me if I'm wrong, but it seems to me the unbuffered channel is safe. 

Tracing both paths:
* Connect succeeds before the timer: `ch <- res` succeeds.
* Timer is fired first: `Done()` is closed and the second case is exercised.

Arc676: I think you're correct. There is no dependency on this being a buffered channel. In any case, all paths reading from and writing to the channel are non-blocking since they are in `select` blocks. I've made the channel unbuffered.

internal/systemd/systemd_test.go
Arc676: This is a good point. I'm not sure what the most meaningful solution is. For the moment, I've added a check such that the tests involving real `systemd` connections are skipped when the host shutdown flag is set to `false` (emulating the behavior in production).

I suppose we could try to use a fake socket, but I'm not sure how to do this efficiently. We'd have to either replace the connection setup function, which makes the test deviate from the production behavior, or emulate the real `systemd` connection's setup and `Hello` calls such that `dbus` thinks that it connected properly. This would mean maintaining a sort of minimal `dbus` server in the test code. Or maybe you have a different idea? I've never used `dbus` in my own projects so I don't know much about using it.

rajathagasthya: > For the moment, I've added a check such that the tests involving real systemd connections are skipped when the host shutdown flag is set to false (emulating the behavior in production).

Nothing in tests or CI sets that env var though, so it doesn't have an effect.

I don't know how to use a fake socket either, to be honest. I asked Claude and it is suggested the following diff (for transparency, I don't fully understand it):

```diff
diff --git a/internal/systemd/systemd_test.go b/internal/systemd/systemd_test.go
index 67c203e9..c83b6936 100644
--- a/internal/systemd/systemd_test.go
+++ b/internal/systemd/systemd_test.go
@@ -18,6 +18,8 @@
 
 import (
 	"context"
+	"errors"
+	"io"
 	"net"
 	"os"
 	"path/filepath"
@@ -68,14 +70,18 @@ func TestNewManagerWithTimeoutUnresponsiveSocket(t *testing.T) {
 
 	// Accept connections but never write a reply, mimicking a bind-mounted host
 	// socket with no systemd/D-Bus daemon behind it.
+	accepted := make(chan net.Conn, 1)
 	go func() {
 		for {
 			conn, err := listener.Accept()
 			if err != nil {
 				return
 			}
-			// Hold the connection open without responding.
 			t.Cleanup(func() { _ = conn.Close() })
+			select {
+			case accepted <- conn:
+			default:
+			}
 		}
 	}()
 
@@ -93,6 +99,25 @@ func TestNewManagerWithTimeoutUnresponsiveSocket(t *testing.T) {
 	if elapsed > 5*time.Second {
 		t.Errorf("connect took %s, expected it to fail fast near the %s timeout", elapsed, timeout)
 	}
+
+	// Cancelling the connection context must close the client socket, so the
+	// server side of the stuck handshake reads EOF.
+	var serverConn net.Conn
+	select {
+	case serverConn = <-accepted:
+	case <-time.After(2 * time.Second):
+		t.Fatal("no connection was accepted")
+	}
+	_ = serverConn.SetReadDeadline(time.Now().Add(2 * time.Second))
+	for {
+		_, err := serverConn.Read(make([]byte, 64))
+		if errors.Is(err, io.EOF) {
+			break
+		}
+		if err != nil {
+			t.Fatalf("expected EOF on the server side after the timeout, got: %v", err)
+		}
+	}
 }
 
 func TestNewManagerWithTimeoutMissingSocket(t *testing.T) {
@@ -125,55 +150,22 @@ func TestManagerCloseNil(t *testing.T) {
 	}
 }
 
-func TestManagerRace(t *testing.T) {
-	if os.Getenv("WITH_SHUTDOWN_HOST_GPU_CLIENTS") == "false" {
-		t.Skip()
-	}
-
-	const timeout = 1 * time.Second
-
-	var conn *dbus.Conn
-
-	delayedConnect := func(ctx context.Context) (*dbus.Conn, error) {
-		var err error
-
-		// Intercept the connection so we can check its state afterwards
-		conn, err = dbus.NewSystemConnectionContext(ctx)
+func TestNewManagerWithTimeoutLateConnector(t *testing.T) {
+	const timeout = 100 * time.Millisecond
 
-		// Only send result after context is canceled to simulate race condition
+	// The connector only returns after the context is cancelled, so its result
+	// can never arrive before the timer fires.
+	connect := func(ctx context.Context) (*dbus.Conn, error) {
 		<-ctx.Done()
-
-		return conn, err
+		return nil, ctx.Err()
 	}
 
-	mgr, err := newManagerWithTimeout(context.Background(), timeout, delayedConnect)
+	start := time.Now()
+	mgr, err := newManagerWithTimeout(context.Background(), timeout, connect)
+	elapsed := time.Since(start)
 
 	expectTimeout(t, mgr, err)
-
-	if conn.Connected() {
-		t.Errorf("expected connection to be dropped after timeout")
-	}
-}
-
-func TestCancelDbus(t *testing.T) {
-	if os.Getenv("WITH_SHUTDOWN_HOST_GPU_CLIENTS") == "false" {
-		t.Skip()
-	}
-
-	ctx, cancel := context.WithCancel(context.Background())
-
-	conn, err := dbus.NewSystemConnectionContext(ctx)
-	if err != nil {
-		t.Errorf("expected successful connection, got error: %v", err)
-	}
-
-	if !conn.Connected() {
-		t.Error("connection is not connected")
-	}
-
-	cancel()
-
-	if conn.Connected() {
-		t.Error("canceling connection context did not disconnect dbus")
+	if elapsed > 2*time.Second {
+		t.Errorf("timeout return took %s, expected it near %s", elapsed, timeout)
 	}
 }
```

Arc676: > Nothing in tests or CI sets that env var though, so it doesn't have an effect.

True, but this is straightforward to change. We could, for example, update the `test` target in the `Makefile` to include

```bash
if which systemctl >&/dev/null && systemctl is-system-running >&/dev/null; then
  export WITH_SHUTDOWN_HOST_GPU_CLIENTS=true
else
  export WITH_SHUTDOWN_HOST_GPU_CLIENTS=false
fi
```

Then the `systemd` tests will be skipped if the test runner doesn't use `systemd`.

---

I asked MS Copilot (GPT) to generate code for a `dbus` listener that would successfully reply to the `Hello` call that DBus uses and keep connections open. This ensures that `dbus.NewSystemConnectionContext` succeeds and `conn.Connected()` returns `true` afterwards. Basically, this emulates a real, successful `systemd` connection, except it replies to any other calls with `UnknownMethod`. I ran the tests on macOS to verify that they pass when `systemd` isn't available.

In the interest of time, I've pushed the code to this branch so you can also take a look. However, I've deliberately omitted the signoff trailer because the generated code is almost 800 lines and I haven't thoroughly reviewed the DBus implementation, nor do I want to without confirmation that it's worth keeping such a massive chunk of code around just for a unit test. Personally, I'd find this a bit excessive.

---

There is one more alternative approach: instead of creating a dummy socket in the Go code, we can just create a dummy DBus listener in the CI pipeline. This is a lot shorter, only adding about 50 lines and nothing that needs extra maintenance to the test. GPT suggests the following:

```go
type TestBus struct {
	cmd     *exec.Cmd
	address string
}

func NewTestBus(t *testing.T) *TestBus {
	t.Helper()

	ctx, cancel := context.WithCancel(context.Background())

	cmd := exec.CommandContext(
		ctx,
		"dbus-daemon",
		"--session",
		"--nofork",
		"--print-address=1",
	)

	stdout, err := cmd.StdoutPipe()
	if err != nil {
		cancel()
		t.Fatal(err)
	}

	if err := cmd.Start(); err != nil {
		cancel()
		t.Fatal(err)
	}

	r := bufio.NewReader(stdout)

	address, err := r.ReadString('\n')
	if err != nil {
		cancel()
		_ = cmd.Wait()
		t.Fatal(err)
	}

	bus := &TestBus{
		cmd:     cmd,
		address: strings.TrimSpace(address),
	}

	t.Setenv("DBUS_SYSTEM_BUS_ADDRESS", bus.address)

	t.Cleanup(func() {
		cancel()
		_ = cmd.Wait()
	})

	return bus
}
```

This also works and is much shorter, but introduces a different dependency to the tests: this doesn't run on macOS anymore because there is no `dbus-daemon`. Of course, most systems on which the tests are likely to run will probably have this binary. But the same could be said about `systemd` itself. For example, Talos has neither `systemd` nor the `dbus-daemon` binary. I suppose it's slightly safer to assume that `dbus-daemon` is present since even systems without `systemd` might still have DBus for other purposes.

If I had to choose, I suppose I'd rather spawn a subshell and depend on an external binary to simulate the `systemd` DBus connection than add a partial DBus implementation to the test code. Let me know what you think.

sylvesterkaczmarek:
I would avoid keeping the ~800-line hand-written D-Bus server just for these tests. The production ownership/timeout logic is unchanged and my earlier leak concern is already resolved.

For the remaining portability concern, I would prefer one of two smaller options:
1. Keep the deterministic injected-connector tests for newManagerWithTimeout and drop the test that is really asserting go-systemd's own cancellation behaviour; or
2. If an integration-level D-Bus cancellation check is valuable, launch dbus-daemon in a small test helper and skip that specific test when the binary is unavailable.

I lean toward the first option because it keeps this PR focused on mig-parted's logic rather than maintaining a partial D-Bus implementation.

internal/systemd/systemd_test.go
rajathagasthya: @Arc676 Just stepping back a bit — would it make sense to just remove `TestManagerRace` and `TestCancelDbus`? They exist because of the `dbus` connection leak concern, which you've established doesn't occur. I don't think it's worth all this machinery.

Arc676: I agree that the machinery is a bit overcomplicated. Do you think it's worth keeping some kind of test to ensure that the connection leak concern is unfounded? We've established that `dbus` handles the connection closure properly but is it enough to know this now and not continue testing for it in the future? I guess if we trust the `dbus` implementation to not change this particular detail, then we don't really need to go out of our way to verify it. I'm OK with just leaving in a comment addressing this issue and removing the tests.

rajathagasthya: > Do you think it's worth keeping some kind of test to ensure that the connection leak concern is unfounded?

Sure, let's do it. I think your dummy dbus listener should work. We can ensure the test is skipped on MacOS. The following (crafted with Claude's help) is probably sufficient coverage:

```go
// startTestBus runs a private dbus-daemon and points the bus address at it.
// Skips the test when the binary is not installed.
func startTestBus(t *testing.T) {
	t.Helper()
	if _, err := exec.LookPath("dbus-daemon"); err != nil {
		t.Skip("dbus-daemon not found")
	}

	socketPath := shortSocketPath(t)
	ctx, cancel := context.WithCancel(context.Background())

	//nolint:gosec // fixed binary name; only the socket path varies
	cmd := exec.CommandContext(ctx, "dbus-daemon", "--session", "--nofork", "--print-address=1",
		"--address=unix:path="+socketPath)
	stdout, err := cmd.StdoutPipe()
	if err != nil {
		cancel()
		t.Fatal(err)
	}
	if err := cmd.Start(); err != nil {
		cancel()
		t.Fatal(err)
	}
	t.Cleanup(func() {
		cancel()
		_ = cmd.Wait()
	})

	// The daemon prints its address once it is listening.
	if _, err := bufio.NewReader(stdout).ReadString('\n'); err != nil {
		t.Fatalf("dbus-daemon did not report an address: %v", err)
	}
	setBusAddress(t, socketPath)
}

func TestCancelDbus(t *testing.T) {
	startTestBus(t)

	ctx, cancel := context.WithCancel(context.Background())
	conn, err := dbus.NewSystemConnectionContext(ctx)
	if err != nil {
		t.Fatalf("expected successful connection, got error: %v", err)
	}
	if !conn.Connected() {
		t.Fatal("not connected")
	}

	cancel()

	if conn.Connected() {
		t.Error("canceling connection context did not disconnect dbus")
	}
}
```

Arc676: Updated the branch with the binary-based dummy-DBus approach and also squashed the commits.

internal/systemd/systemd_test.go
Arc676: Switched to dummy DBus implementation.

sylvesterkaczmarek (approved):
Rechecked current d774ee01. The lazy-init path is now clean: New() does not touch D-Bus, cleanup() only closes an already-cached manager, and the timeout worker owns any late result after cancellation. The earlier test-portability concern is also resolved by using a small real dbus-daemon helper that skips when unavailable rather than maintaining a hand-written D-Bus server. No remaining blocker from my review.

PR #408, opened 2026-06-24
Start driver services non-blocking during boot to avoid deadlock

rajathagasthya:
`nvidia-mig-manager.service` runs `Before=nvidia-gpu-reset.target`, while the
driver services it restarts (nvsm, nvidia-dcgm) are ordered After that
target. The `apply-exit` hook started them with a blocking `systemctl
start`, which deadlocks during boot: the service waits for the start to
finish, but the target -- and therefore those services -- cannot start
until the service itself completes.

Before commit db06e54e, service.sh did not default MIG_PARTED_HOOKS_FILE, so under
systemd the hooks were silently skipped and apply-exit never started the
driver services. Once that fix landed (v0.14.0), apply-exit began running
`systemctl start nvsm.service` during boot, which exposed the ordering
deadlock.

Use `systemctl is-system-running` to detect that the system is still
starting and enqueue the start with `--no-block` in that case, letting
the service finish so systemd brings the driver services up once the
target is reached. Once the system is running (a runtime reconfigure),
start synchronously as before so the apply still waits for the services
and reports start failures.

### DIFF
--- a/deployments/systemd/hooks.sh
+++ b/deployments/systemd/hooks.sh
@@ -92,7 +92,21 @@ function stop_driver_services() {
 }
 
 function start_driver_services() {
-	nvidia-mig-manager::service::start_systemd_services driver_services
+	# Driver services are ordered After=nvidia-gpu-reset.target, and this
+	# service runs Before that target. During boot, starting them
+	# synchronously here deadlocks: this service waits on them while the
+	# target (and therefore those services) waits on this service to finish.
+	# While the system is still starting up, enqueue the start without
+	# waiting; systemd starts them once this service completes and the target
+	# is reached. Once the system is running, start synchronously as before so
+	# a runtime reconfigure still waits for the services and reports failures.
+	local start_args=""
+	local state
+	state="$(systemctl is-system-running 2>/dev/null)"
+	if [ "${state}" != "running" ] && [ "${state}" != "degraded" ]; then
+		start_args="--no-block"
+	fi
+	nvidia-mig-manager::service::start_systemd_services driver_services "${start_args}"
 	return ${?}
 }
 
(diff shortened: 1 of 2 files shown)


deployments/systemd/hooks.sh
```diff
@@ -92,7 +92,21 @@ function stop_driver_services() {
...
+	# synchronously here deadlocks: this service waits on them while the
+	# target (and therefore those services) waits on this service to finish.
+	# While the system is still starting up, enqueue the start without
+	# waiting; systemd starts them once this service completes and the target
+	# is reached. Once the system is running, start synchronously as before so
+	# a runtime reconfigure still waits for the services and reports failures.
+	local start_args=""
+	local state
+	state="$(systemctl is-system-running 2>/dev/null)"
+	if [ "${state}" != "running" ] && [ "${state}" != "degraded" ]; then
```
cdesiniotis, line 106: Question -- what does `degraded` mean in this context? And why are we starting the services synchronously when the system is in this state too?

rajathagasthya: `degraded` is what is reported once boot has finished, but at least one unit has failed. It still means the system is fully operational. 

In our case, the deadlock only exists during boot when `nvidia-gpu-reset.target` hasn't been reached yet. By the time the system is `running` or `degraded`, the boot has completed and `nvidia-gpu-reset.target` is active (meaning `nvidia-mig-manager.service` is running successfully). We're still starting services synchronously when `degraded` to surface start failures when a user does a manual `nvidia-mig-parted apply`.

cdesiniotis: Got it, thanks for the details here!

cdesiniotis (approved):

PR #422, opened 2026-07-02
add oneshot services  timeout

### LINKED ISSUE #68: oneshot services causes boot failures due to lack of timeout
`nvidia-mig-manager.service` is `Type=oneshot`.
`DefaultTimeoutStartSec` is not used for oneshot services, which causes the entire system to fail to boot stuck for nvidia-mig-manager.service to complete.

Boot failure is worse than a failed / degraded service.

A `TimeoutStartSec` should be added to this to at least allow the system to boot in a degraded state (for debug / recovery without OOB BMC / IPMI / KVM).

The root cause may be #11 , but a timeout addition will make this more resilient.

image

shivakunv:
Issue: [68](https://github.com/NVIDIA/mig-parted/issues/68) The nvidia-mig-manager systemd service is configured as Type=oneshot, meaning systemd waits for the service to complete before proceeding. The MIG partitioning operation can take longer than the default systemd startup timeout, causing the service to be killed prematurely and fail. GPU nodes with time-consuming MIG operations would silently fail at boot, leaving GPUs in an inconsistent state

Fix: This PR adds TimeoutStartSec=480 to deployments/systemd/nvidia-mig-manager.service, setting a 8 minute startup timeout so the oneshot service has sufficient time to complete MIG reconfiguration. If it exceeds that timeout window, systemd marks the service as failed and boots the system in a degraded state, giving the user an SSH shell to investigate.

### DIFF
--- a/deployments/systemd/nvidia-mig-manager.service
+++ b/deployments/systemd/nvidia-mig-manager.service
@@ -20,6 +20,7 @@ Before=nvidia-gpu-reset.target
 
 [Service]
 Type=oneshot
+TimeoutStartSec=900
 ExecStart=/bin/bash /etc/nvidia-mig-manager/service.sh
 
 [Install]
--- a/deployments/systemd/nvidia-mig-manager.service
+++ b/deployments/systemd/nvidia-mig-manager.service
@@ -20,6 +20,7 @@ Before=nvidia-gpu-reset.target
 
 [Service]
 Type=oneshot
+TimeoutStartSec=480
 ExecStart=/bin/bash /etc/nvidia-mig-manager/service.sh
 
 [Install]


rajathagasthya (approved):

tariq1890:
Can we add a description to this PR?

shivakunv:
> Can we add a description to this PR?

added

karthikvetrivel (approved):

rahulait (approved):

tariq1890:
@rajathagasthya  and I discussed this offline. The `900` second timeout may be too long as it could elongate boot times by a lot. To reach a middle ground, let's start with `480` seconds as the timeout. If we find that it's not long enough of a timeout, we can then proceed to increase the timeout seconds.

deployments/systemd/nvidia-mig-manager.service
```diff
@@ -20,6 +20,7 @@ Before=nvidia-gpu-reset.target
 
 [Service]
 Type=oneshot
+TimeoutStartSec=900
```
tariq1890:
```suggestion
TimeoutStartSec=480
```

rajathagasthya (approved):

tariq1890 (approved):

PR #426, opened 2026-07-07
add support for bouncing third-party gpu client pods during mig config apply

tariq1890:
This PR is linked to https://github.com/NVIDIA/gpu-operator/pull/2607.

This adds the ability for the `mig-manager` to detect third-party/custom gpu client pods during MIG config updates and to evict them as they too will be holding onto gpu device handles. With this change, users can deploy their custom gpu client pods and also take advantage of the automatic toggling of gpu client components (just as what is done for the first-class gpu clients namely the "operands").

NOTE: The third-party gpu client pods MUST add the nvidia.com/gpu.deploy.client: true label to its nodeSelector so that the `mig-manager` detects and performs the automated toggle during a MIG config change.

### DIFF
--- a/pkg/mig/reconfigure/reconfigure.go
+++ b/pkg/mig/reconfigure/reconfigure.go
@@ -27,7 +27,9 @@ import (
 	"time"
 
 	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
@@ -42,6 +44,7 @@ const (
 	dcgmExporterDeployLabel        = "nvidia.com/gpu.deploy.dcgm-exporter"
 	dcgmDeployLabel                = "nvidia.com/gpu.deploy.dcgm"
 	nvsmDeployLabel                = "nvidia.com/gpu.deploy.nvsm"
+	gpuClientDeployLabel           = "nvidia.com/gpu.deploy.client"
 
 	migStateSuccess = "success"
 	migStateFailed  = "failed"
@@ -93,6 +96,7 @@ type Reconfigure struct {
 	dcgmExporterDeployed  string
 	dcgmDeployed          string
 	nvsmDeployed          string
+	gpuClientDeployed     string
 	currentState          string
 	migModeChangeRequired bool
 	stoppedServices       []string
@@ -255,6 +259,7 @@ func (r *Reconfigure) getCurrentNodeLabels() error {
 	r.dcgmExporterDeployed = labels[dcgmExporterDeployLabel]
 	r.dcgmDeployed = labels[dcgmDeployLabel]
 	r.nvsmDeployed = labels[nvsmDeployLabel]
+	r.gpuClientDeployed = labels[gpuClientDeployLabel]
 	r.currentState = labels[migConfigStateLabel]
 
 	log.Infof("Current value of '%s=%s'\n", devicePluginDeployLabel, r.pluginDeployed)
@@ -262,6 +267,8 @@ func (r *Reconfigure) getCurrentNodeLabels() error {
 	log.Infof("Current value of '%s=%s'\n", dcgmExporterDeployLabel, r.dcgmExporterDeployed)
 	log.Infof("Current value of '%s=%s'\n", dcgmDeployLabel, r.dcgmDeployed)
 	log.Infof("Current value of '%s=%s'\n", nvsmDeployLabel, r.nvsmDeployed)
+	log.Infof("Current value of '%s=%s'\n", gpuClientDeployLabel, r.gpuClientDeployed)
+
 	log.Infof("Current value of '%s=%s'\n", migConfigStateLabel, r.currentState)
 
 	return nil
@@ -371,6 +378,9 @@ func (r *Reconfigure) shutdownKubernetesGPUClients() error {
 		dcgmDeployLabel:                r.maybeSetPaused(r.dcgmDeployed),
 		nvsmDeployLabel:                r.maybeSetPaused(r.nvsmDeployed),
 	}
+	if len(r.gpuClientDeployed) > 0 {
+		labels[gpuClientDeployLabel] = r.maybeSetPaused(r.gpuClientDeployed)
+	}
 
 	return r.setNodeLabels(labels)
 }
@@ -397,6 +407,11 @@ func (r *Reconfigure) waitForPodsToBeDeleted() error {
 	log.Info("Waiting for dcgm to shutdown")
 	if err := r.waitForPodDeletion("app=nvidia-dcgm", timeout); err != nil {
 		return fmt.Errorf("dcgm pod did not shutdown: %w", err)
+	}
+
+	log.Infof("Waiting for any daemon set pods with nodeSelector key %s to shutdown", gpuClientDeployLabel)
+	if err := r.waitForPodsWithNodeSelector(gpuClientDeployLabel, timeout); err != nil {
+		return fmt.Errorf("third-party gpu client pods did not shutdown: %w", err)
 	}
 
 	log.Info("Removing the cuda-validator pod")
@@ -508,6 +523,9 @@ func (r *Reconfigure) restartKubernetesGPUClients() error {
 		dcgmDeployLabel:                r.maybeSetTrue(r.dcgmDeployed),
 		nvsmDeployLabel:                r.maybeSetTrue(r.nvsmDeployed),
 	}
+	if len(r.gpuClientDeployed) > 0 {
+		labels[gpuClientDeployLabel] = r.maybeSetTrue(r.gpuClientDeployed)
+	}
 
 	return r.setNodeLabels(labels)
 }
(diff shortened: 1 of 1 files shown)


pkg/mig/reconfigure/reconfigure.go
```diff
@@ -370,6 +377,7 @@ func (r *Reconfigure) shutdownKubernetesGPUClients() error {
 		dcgmExporterDeployLabel:        r.maybeSetPaused(r.dcgmExporterDeployed),
 		dcgmDeployLabel:                r.maybeSetPaused(r.dcgmDeployed),
 		nvsmDeployLabel:                r.maybeSetPaused(r.nvsmDeployed),
+		gpuClientDeployLabel:           r.maybeSetPaused(r.gpuClientDeployed),
```
rajathagasthya: I noticed the corresponding k8s-device-plugin has a label existence check in  https://github.com/NVIDIA/k8s-driver-manager/pull/205/changes#diff-34f0af81ba61aa2b1ce6de2801e6cd7c160856c3466c57212bece1519403734cR573-R575, but this does not. Is it intentional?

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -507,6 +520,7 @@ func (r *Reconfigure) restartKubernetesGPUClients() error {
 		dcgmExporterDeployLabel:        r.maybeSetTrue(r.dcgmExporterDeployed),
 		dcgmDeployLabel:                r.maybeSetTrue(r.dcgmDeployed),
 		nvsmDeployLabel:                r.maybeSetTrue(r.nvsmDeployed),
+		gpuClientDeployLabel:           r.maybeSetTrue(r.gpuClientDeployed),
```
rajathagasthya: Same comment about label existence check.

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -538,6 +552,7 @@ func (r *Reconfigure) setState(state string) error {
 			dcgmExporterDeployLabel:        r.maybeSetTrue(r.dcgmExporterDeployed),
 			dcgmDeployLabel:                r.maybeSetTrue(r.dcgmDeployed),
 			nvsmDeployLabel:                r.maybeSetTrue(r.nvsmDeployed),
+			gpuClientDeployLabel:           r.maybeSetTrue(r.gpuClientDeployed),
```
rajathagasthya: Same comment about label existence check.

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -621,6 +636,69 @@ func (r *Reconfigure) deletePod(labelSelector string) error {
...
+//
+// Owner controller lookups are cached by UID across poll iterations. We don't expect the owners to actually change
+// during the lifetime of the wait, so caching the owners list optimises for kube API call volumes.
+func (r *Reconfigure) waitForPodsWithParentNodeSelector(nodeSelectorKey string, timeout time.Duration) error {
+	// ownerMatchCache maps owner UID → whether its nodeSelector contains nodeSelectorKey.
+	// Owner controllers are stable during the wait; only the pods themselves are terminating.
+	ownerMatchCache := make(map[types.UID]bool)
+
+	return wait.PollUntilContextTimeout(r.ctx, 5*time.Second, timeout, true, func(ctx context.Context) (bool, error) {
+		podList, err := r.clientset.CoreV1().Pods(corev1.NamespaceAll).List(ctx, metav1.ListOptions{
```
rajathagasthya, line 650: We're listing pods across all namespaces, so we will need a `pods list` permission in the ClusterRole too. I only see `daemonsets get` in https://github.com/NVIDIA/gpu-operator/pull/2607.

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -621,6 +636,69 @@ func (r *Reconfigure) deletePod(labelSelector string) error {
...
+		if err != nil {
+			return false, fmt.Errorf("failed to list pods on node %s: %w", r.opts.NodeName, err)
+		}
+
+		matchCount := 0
+		for _, pod := range podList.Items {
+			for _, ownerRef := range pod.OwnerReferences {
+				matches, ok := ownerMatchCache[ownerRef.UID]
+				if !ok {
+					nodeSelector, err := r.getOwnerNodeSelector(ctx, pod.Namespace, ownerRef)
```
rajathagasthya: The pod already has the node selector in `pod.Spec.nodeSelector`. Is it necessary to get the value from the owner daemonset? The value will be the same as it's just copied from the daemonset template anyway.

We can just move up the daemonset owner check here. Something like this:

```go
for _, pod := range podList.Items {
    if _, ok := pod.Spec.NodeSelector[nodeSelectorKey]; !ok {
        continue
    }
    owner := metav1.GetControllerOf(&pod)
    if owner != nil && owner.Kind == "DaemonSet" {
        matchCount++
    } else {
        // log skipped pod
    }
    // rest of the code
}
```
If we do this, we don't need to add `daemonsets get` permission either, just `pods list`.

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -399,6 +407,11 @@ func (r *Reconfigure) waitForPodsToBeDeleted() error {
 		return fmt.Errorf("dcgm pod did not shutdown: %w", err)
 	}
 
+	log.Infof("Waiting for any pods with parent nodeSelector key %s to shutdown", gpuClientDeployLabel)
+	if err := r.waitForPodsWithParentNodeSelector(gpuClientDeployLabel, timeout); err != nil {
+		return fmt.Errorf("third-party gpu client pods did not shutdown: %w", err)
```
rajathagasthya, line 414: Should we set `r.setState(migStateFailed)` if this fails?

pkg/mig/reconfigure/reconfigure.go
rajathagasthya: Oh nevermind, I saw it was just added there.

pkg/mig/reconfigure/reconfigure.go
tariq1890: Good catch.

tariq1890: I'll fix this

pkg/mig/reconfigure/reconfigure.go
tariq1890: You're right. The `daemonsets get` is unnecessary. It seems like pods that are parent'ed by daemonsets will have a nodeSelector in the PodSpec

pkg/mig/reconfigure/reconfigure.go
tariq1890: We don't do this for the other labels. If we want to add this, I can revisit it another PR

tariq1890: This is the method that calls all of these methods. So we could set the failed state here

```
	if err := r.waitForPodsToBeDeleted(); err != nil {
		return fmt.Errorf("failed to wait for pods to be deleted: %w", err)
	}
```

But yes, I think we can revisit this in another PR

pkg/mig/reconfigure/reconfigure.go
tariq1890: Actually, if you look at the label above , `nvsmDeployLabel`
```
nvsmDeployLabel                = "nvidia.com/gpu.deploy.nvsm"
```

This does not exist in most gpu clusters; gpu-operator does not apply this label. So far, mig-manager has been working fine without a label existence check for this one.

tariq1890:
@rajathagasthya I have addressed most (if not all) of your review comments. Please check now

pkg/mig/reconfigure/reconfigure.go
rajathagasthya: Thanks, a follow up is fine.

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -621,6 +643,35 @@ func (r *Reconfigure) deletePod(labelSelector string) error {
...
+		if err != nil {
+			return false, fmt.Errorf("failed to list pods on node %s: %w", r.opts.NodeName, err)
+		}
+
+		matchCount := 0
+		for _, pod := range podList.Items {
+			ownerRef := metav1.GetControllerOf(&pod)
+			if ownerRef == nil || ownerRef.Kind != "DaemonSet" {
+				continue
+			}
```
rajathagasthya, line 662: nit: you could add a warning log here if a pod without a daemonset owner has this nodeSelector expecting an eviction, so there's some visibility into why it wasn't evicted.

If you do decide to do it, it's better to use a `map[types.UID]bool` to keep track of visited non-daemonset owner pods, so the same log line isn't repeated on every poll.

pkg/mig/reconfigure/reconfigure.go
```diff
@@ -399,6 +409,11 @@ func (r *Reconfigure) waitForPodsToBeDeleted() error {
 		return fmt.Errorf("dcgm pod did not shutdown: %w", err)
 	}
 
+	log.Infof("Waiting for any daemon set pods with nodeSelector key %s to shutdown", gpuClientDeployLabel)
+	if err := r.waitForPodsWithNodeSelector(gpuClientDeployLabel, timeout); err != nil {
```
rahulait, line 413: Same here as k8s-driver-manager, we should respect if someone has set:
```
nvidia.com/gpu.deploy.client: false
```
We should not wait for eviction in this case. I am expecting users will toggle this label more than deleting the label altogether if they don't want their pod to be auto-evicted.

pkg/mig/reconfigure/reconfigure.go
tariq1890: I'd rather avoid it tbh. I'd prefer mentioning it in the documentation. Maintaining a map for this warning log seems like too much effort for too little gain

rajathagasthya: Fine with me.

rajathagasthya (approved):

PR #428, opened 2026-07-09
Restore /var/run symlink to /run in the container image

### LINKED ISSUE #429: [Bug]: GPU reset fails in images built from main: /var/run -> /run symlink was dropped
**Describe the bug**
`nvidia-smi` discovers the `nvidia-persistenced` socket at `/var/run/nvidia-persistenced/socket`, while the container toolkit injects the socket into the container at `/run/nvidia-persistenced/socket`. Commit 3944439b added a `/var/run -> /run` symlink to the image so `nvidia-smi` can reach the socket and coordinate with persistenced when resetting GPUs.

The switch to the non-dev distroless base image in #385 (19fe9af0) dropped that symlink, and the new base ships `/var/run` as a plain empty directory. As a result, `nvidia-smi -r` (invoked by `nvidia-mig-parted apply` when a MIG mode change is pending) cannot see the persistenced socket, and the reset fails with the GPU reported as in use by another client (`nvidia-persistenced`).

Only images built from `main` after 19fe9af0 are affected. Released versions (v0.13.0 through v0.14.3) contain the symlink and are unaffected.

**To Reproduce**
1. Build the mig-manager container image from `main` (any commit after 19fe9af0).
2. Deploy it on a node where `nvidia-persistenced` is running (e.g. the standard GPU Operator driver container setup).
3. Label the node with a MIG config that requires a MIG mode change.
4. The reconfigure fails at the GPU reset step: `nvidia-smi -r` reports the GPU is in use by another client.

**Expected behavior**
The GPU reset succeeds: `nvidia-smi` reaches the persistenced socket via `/var/run/nvidia-persistenced/socket` and coordinates the reset, as it does in v0.14.x images.

**Environment (please provide the following information):**
 - `mig-parted` version: `main` (after 19fe9af0, unreleased)
 - Host OS: any
 - Kernel Version: any
 - NVIDIA Driver Version: any
 - GPU Model(s): any MIG-capable Ampere GPUs (Hopper+ does not require a GPU reset)

Fix in #428.

rajathagasthya:
nvidia-smi discovers the nvidia-persistenced socket at
`/var/run/nvidia-persistenced/socket`, while the container toolkit injects
it at `/run/nvidia-persistenced/socket`. Commit 3944439b symlinked
`/var/run to /run` so nvidia-smi can reach the socket and coordinate with
persistenced when resetting GPUs. The switch to the non-dev distroless
base in 19fe9af0 dropped that symlink, and the new base ships `/var/run`
as a plain empty directory, so GPU resets after a MIG mode change fail
with the GPU reported as in use by another client (nvidia-persistenced).

Recreate the symlink in the final image stage using the busybox applets,
since the non-dev base image has no shell of its own.

Fixes #429

### DIFF
--- a/deployments/container/Dockerfile
+++ b/deployments/container/Dockerfile
@@ -72,6 +72,10 @@ USER 0:0
 
 COPY --from=shell /busybox /busybox
 RUN ["/busybox/ln", "-s", "/busybox/sh", "/bin/sh"]
+# nvidia-smi looks for the persistenced socket at /var/run/..., but the
+# container toolkit injects it at /run/... Symlink /var/run to /run so GPU
+# resets can use nvidia-persistenced.
+RUN ["/busybox/sh", "-c", "/busybox/rm -r /var/run && /busybox/ln -s /run /var/run"]
 ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/busybox
 
 COPY --from=build /artifacts/nvidia-mig-parted  /usr/bin/nvidia-mig-parted


tariq1890 (approved):

tariq1890:
Validated the image built from this branch - `ghcr.io/nvidia/k8s-mig-manager:fafbbc5e`. The mig-manager is now able to perform GPU resets on A100 nodes successfully

PR #431, opened 2026-07-09
fix: prevent the MIG reconfigure from hanging on systemd-less hosts

lexfrei:
## What

On hosts without systemd — notably Talos Linux — `k8s-mig-manager` v0.14.0 hangs indefinitely when the `nvidia.com/mig.config` label changes, and never applies the new MIG geometry.

The Go-native reconfigure path opens a systemd D-Bus connection unconditionally when it is constructed. On a systemd-less host the system bus socket (`/run/dbus/system_bus_socket`) still exists — it is bind-mounted from the host — but nothing answers it, so the D-Bus auth handshake performs a blocking read that never returns. This happens even when `WITH_SHUTDOWN_HOST_GPU_CLIENTS=false`, i.e. when no host-systemd work is actually required.

## Why it matters

Most of the reconfigure flow does not need systemd at all — the MIG geometry change is applied by the `nvidia-mig-parted` binary, not through systemd. systemd is only needed to persist the host mig-manager state file (a daemon-reload) and to stop/restart host GPU client services. A host that never touches those paths should never need a D-Bus connection.

## Changes

- Connect to systemd D-Bus lazily. Constructing the reconfigurer no longer dials D-Bus; the connection is established (and cached) the first time a code path genuinely needs systemd. On the common systemd-less path the socket is never dialed, so the reconfigure completes.
- Bound the D-Bus connection setup with a timeout, as a backstop for the paths that do require systemd. If the socket is unresponsive, the connection attempt now fails fast with a clear error instead of blocking forever.

## Testing

- Unit tests added for the connection timeout (fast failure on an unresponsive socket, immediate failure on a missing socket) and for the lazy behavior (construction does not dial; a failed connection is not cached; cleanup is safe when systemd was never used).
- `go build ./...`, `go vet`, `gofmt`, and `golangci-lint` pass on the changed packages.
- End-to-end behavior on an actual systemd-less GPU host (Talos) cannot be exercised in CI; it requires the target hardware and OS.

Addresses #356

### DIFF
--- a/internal/systemd/systemd.go
+++ b/internal/systemd/systemd.go
@@ -27,30 +27,85 @@ import (
 	"github.com/coreos/go-systemd/v22/dbus"
 )
 
+// systemdConnectTimeout bounds how long NewManager waits for the systemd D-Bus
+// connection (dial plus auth handshake) to be established. On hosts without a
+// running systemd/D-Bus daemon the system bus socket may still exist - for
+// example when /run/dbus/system_bus_socket is bind-mounted from the host - but
+// never answer, in which case the underlying auth handshake would otherwise
+// block forever. Bounding it turns that indefinite hang into a fast, actionable
+// error.
+const systemdConnectTimeout = 10 * time.Second
+
 // Manager handles systemd operations using the D-Bus API
 type Manager struct {
 	ctx context.Context
 
-	conn *dbus.Conn
+	conn   *dbus.Conn
+	cancel context.CancelFunc
 }
 
-// NewManager creates a new Manager instance
+// NewManager creates a new Manager instance connected to the systemd system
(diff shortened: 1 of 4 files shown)


rajathagasthya:
@lexfrei Thanks for your contribution! #382 is already trying to address this issue. I would encourage you to look at it and provide feedback since my recommendation there aligns with your approach in this PR.

lexfrei:
Closing in favour of #382, which already carries both of my commits (`1dd5ebc`, `b5a207a`) with authorship intact — thanks @Arc676 for picking them up. That PR is the broader change, so there's no reason to keep this one open beside it.

@rajathagasthya thanks for the pointer. I've left one piece of review feedback over on #382 rather than here.

PR #435, opened 2026-07-13
Support optional subsystem ID matching in device filters

karthikvetrivel:
## Overview

Device filters identify GPUs by their PCI device and vendor IDs (e.g. `0x20B510DE`). Some GPU products share the same device ID and can only be distinguished by their PCI subsystem IDs, which makes it impossible to target one board variant with a MIG config today.

This PR adds optional subsystem ID matching to device filters:

- The device-filter format gains an optional subsystem component: `<device><vendor>[:<subsystem-device><subsystem-vendor>]`, e.g. `0x20B510DE:0x153710DE`.
- Filters without a subsystem component behave exactly as before and match every board variant of a GPU, so all existing configs are unaffected.
- `types.DeviceID` now carries subsystem device/vendor IDs, populated during GPU enumeration from PCIe config space (using the subsystem fields added in go-nvlib v0.12.0).

## Tested

I added unit tests covering the new filter parsing, string formatting, and matching semantics.

I also tested live on an A100 node by creating MIG configs with subsystem-qualified device filters (`0x20B510DE:0x153310DE`). I verified that a config block with the matching subsystem ID was applied (3g.40gb×2), that a config block with a sibling subsystem ID was correctly skipped, and that `nvidia-mig-parted apply` and `assert` behave unchanged with existing primary-only filters.

### DIFF
--- a/pkg/types/device.go
+++ b/pkg/types/device.go
@@ -19,36 +19,124 @@ package types
 import (
 	"fmt"
 	"strconv"
+	"strings"
 )
 
 // DeviceID represents a GPU Device ID as read from a GPUs PCIe config space.
-type DeviceID uint32
+type DeviceID struct {
+	Device          uint16
+	Vendor          uint16
+	SubsystemDevice uint16
+	SubsystemVendor uint16
+	HasSubsystem    bool
+}
 
 // NewDeviceID constructs a new 'DeviceID' from the device and vendor values pulled from a GPUs PCIe config space.
 func NewDeviceID(device, vendor uint16) DeviceID {
-	return DeviceID((uint32(device) << 16) | uint32(vendor))
+	return DeviceID{
+		Device: device,
+		Vendor: vendor,
+	}
+}
+
+// NewDeviceIDWithSubsystem constructs a new 'DeviceID' with subsystem values.
+func NewDeviceIDWithSubsystem(device, vendor, subDevice, subVendor uint16) DeviceID {
+	return DeviceID{
+		Device:          device,
+		Vendor:          vendor,
+		SubsystemDevice: subDevice,
+		SubsystemVendor: subVendor,
+		HasSubsystem:    true,
+	}
+}
+
+// NewDeviceIDFromPacked constructs a 'DeviceID' from a combined 32-bit device+vendor
+// value, as reported in a GPUs PCIe config space or by NVML.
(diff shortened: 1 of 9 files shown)


rahulait (approved):
LGTM. I would wait for a review from someone else as well before merging this in.

pkg/types/device.go
```diff
@@ -19,36 +19,124 @@ package types
...
+
+// Matches checks if a hardware GPU matches the DeviceID filter.
+// If the filter has a subsystem defined, it requires an exact match on all 4 components.
+// Otherwise, it only matches on the primary device and vendor IDs.
+func (filter DeviceID) Matches(hardware DeviceID) bool {
+	if filter.Device != hardware.Device || filter.Vendor != hardware.Vendor {
+		return false
+	}
+
+	if filter.HasSubsystem {
```
rajathagasthya, line 130: What is the matching behavior when `subsystem_vendor/subsystem_device` is unknown and `go-nvlib` defaults them to `0` since it's just logging a warning currently?

In this case, the device ID construction in `util/device.go` will produce an object like:
```go
{Device: 0x25B6, Vendor: 0x10DE, SubsystemDevice: 0, SubsystemVendor: 0, HasSubsystem: true}
```
It will have `HasSubsystem: true` and `0` subsystem values. And in this method, we're only checking `filter.HasSubsystem` and not `hardware.HasSubsystem`. 

I think a couple of test cases covering this scenario would help. 

```go
{
    name:     "subsystem filter does not match hardware with unknown subsystem",
    filter:   NewDeviceIDWithSubsystem(0x25B6, 0x10DE, 0x14A9, 0x10DE),
    hardware: NewDeviceIDWithSubsystem(0x25B6, 0x10DE, 0x0, 0x0),
    want:     false,
},
{
    name:     "zero subsystem filter matches hardware with unknown subsystem",
    filter:   NewDeviceIDWithSubsystem(0x25B6, 0x10DE, 0x0, 0x0),
    hardware: NewDeviceIDWithSubsystem(0x25B6, 0x10DE, 0x0, 0x0),
    want:     true,
}
```

rajathagasthya (commented):
Minor comment. Overall LGTM.

pkg/types/device.go
karthikvetrivel: Thanks for flagging this — I added these test cases.

rajathagasthya (approved):

PR #449, opened 2026-08-07
Skip container runtime calls in hooks to avoid deadlocks

rajathagasthya:
The `pre-apply-config` hook kills GPU containers by running `docker` and `ctr` unconditionally. Such a call connects to the runtime's socket, and that makes systemd activate the unit behind it when it is not already running. On DGX hosts `docker.service` is ordered after `nvidia-gpu-reset.target`, and `nvidia-mig-manager.service` runs before that target, so an invocation from the hook waits on a start job that cannot run until the hook returns. The service then stays in "activating" indefinitely: `Type=oneshot` disables the start timeout by default and nothing else times out the call.

Call docker/containerd only when it is present and its service is already active. If it is not running, there are no containers to kill. Every call is bounded with timeout.

### DIFF
--- a/deployments/systemd/utils.sh
+++ b/deployments/systemd/utils.sh
@@ -14,6 +14,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+cmd_timeout="${cmd_timeout:-30}"
+
+function nvidia-mig-manager::service::runtime_available() {
+	local client="${1}"
+	local unit="${2}"
+	if ! command -v "${client}" > /dev/null 2>&1; then
+		echo "Skipping container cleanup via ${client}: ${client} is not installed" >&2
+		return 1
+	fi
+	systemctl -q is-active "${unit}"
+	if [ "${?}" != "0" ]; then
+		echo "Skipping container cleanup via ${client}: ${unit} is not active" >&2
+		return 1
+	fi
+	return 0
+}
+
 function nvidia-mig-manager::service::reverse_array() {
 	# first argument is the array to reverse
 	# second is the reversed array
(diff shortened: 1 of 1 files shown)


tariq1890 (approved):

deployments/systemd/utils.sh
```diff
@@ -14,6 +14,31 @@
...
+# Seconds to wait for any single container runtime call before giving up.
+cmd_timeout="${cmd_timeout:-30}"
+
+# Reports whether a container runtime can be queried from a hook without
+# blocking. A client call connects to the runtime's socket, and systemd
+# socket-activates the unit behind it if it is not already running. On DGX
+# hosts docker.service is ordered after nvidia-gpu-reset.target, which is
+# ordered after nvidia-mig-manager.service, creating a deadlock. A runtime
+# whose service is not already active has no running containers to kill, so
+# skipping it is correct.
```
tariq1890: I think is better places in the PR description or the commit message. Let's be mindful of the comments we add to our scripts and code.

deployments/systemd/utils.sh
```diff
@@ -14,6 +14,31 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Seconds to wait for any single container runtime call before giving up.
```
tariq1890: This comment is unnecessary. The variable name is self-explanatory

deployments/systemd/utils.sh
rajathagasthya: Ack. Removed.

deployments/systemd/utils.sh
rajathagasthya: Removed.

rajathagasthya:
Merging using prior approval as only comments have changed since.

PR #450, opened 2026-08-07
Add a THIRD_PARTY_NOTICES.md generator

abrarshivani:
Adds `THIRD_PARTY_NOTICES.md` and the generator that produces it, so the
licenses of the Go deps shipping in `k8s-mig-manager` and in the deb/rpm/tarball
packages are listed in the repo.

- `make third-party-notices` regenerates it
- `make check-third-party-notices` regenerates and diffs it, and runs in CI on every build
- Covers 51 runtime index rows and 26 behind the bundled `nvidia-ctk` binary

## What to review

94% of the diff is generated (10,997 of 11,649 added lines). Hand-written:

| File | Lines |
|---|---:|
| `hack/generate-third-party-notices.sh` | 457 |
| `.github/workflows/golang.yaml` | 27 |
| `Makefile` | 18 / -1 |
| `.gitignore` | 2 |
| `deployments/devel/tools.go` | 1 |

Generated: `THIRD_PARTY_NOTICES.md` (10,997 lines), `deployments/devel/go.mod`
(14) and `go.sum` (133 / -1). The `go mod tidy` only added requirements, no
version moved, and `hack/container-toolkit-version.sh` still reports the same
`nvidia-container-toolkit` version.

## How it works

1. Install the pinned `go-licenses` into `./bin` from `deployments/devel/go.mod`
2. Check the platform list against `DOCKER_BUILD_PLATFORM_OPTIONS` in
   `deployments/container/multi-arch.mk`, fail if out of sync
3. Runtime pass: `go-licenses save` and `csv` over `./cmd/...` with
   `GOFLAGS=-mod=vendor`, once per released platform, then merge (it resolves
   only the host platform and has no union mode)
4. Bundled-binary pass: the same for `nvidia-ctk`, which is pinned in
   `deployments/devel/tools.go` and copied into the released image, resolved
   from the module cache
5. Collapse to one row per package, joining every distinct license
6. Attach `module@version` from `vendor/modules.txt` by longest-prefix match
7. Write header, an index table per surface and the license texts to a temp file,
   then move it into place

The check runs as the `third-party-notices` job in `golang.yaml`, so it
inherits that workflow's triggers rather than declaring its own.

## Implementation notes

- **Only the local module goes to `--ignore`.** It matches raw string prefixes,
  not path segments: passing stdlib names adds the token `go`, which silently
  drops everything under `golang.org/`, `google.golang.org/`, `gopkg.in/` and
  `go.yaml.in/`.
- **Licenses are joined, not picked.** `go-licenses` emits a row per recognized
  license, so key-only dedup would hide `filepath-securejoin`'s MPL-2.0 behind
  its BSD-3-Clause.
- **Rows name `module@version`, not a URL.** In vendor mode `go-licenses` links
  into this repo at `HEAD`, which stops describing released content once `main`
  moves.
- **Output must match on every machine**, since CI diffs it. `LC_ALL=C` on every
  sort and grep. Whole-line `sort -u`, since key-only dedup keeps a different row
  on BSD and GNU sort. The awk counts instead of testing `pkg in lic`, which mawk
  and busybox awk get wrong. `grep -a` so a NUL byte cannot change the measured
  code-fence width.
- **Moved into place with `mv`, not `cp`.** `cp` truncates the target first, so
  an interrupted run would leave a half-written file. The temp file sits next to
  the destination, keeping the rename on one filesystem.
- **`bin/go-licenses` depends on `deployments/devel/go.{mod,sum}`.** Otherwise
  bumping the pin leaves a stale binary here while CI builds the new one, and the
  only symptom is a stale-notices failure that does not reproduce locally.
- **An unidentified license fails the run.** `go-licenses` reports a license it
  cannot classify as `Unknown` and still exits 0, so without a guard the file
  would ship an entry that attributes nothing. Licenses are joined, so the check
  is anchored on both sides: it catches `BSD-3-Clause / Unknown` as well as a
  bare `Unknown`, and leaves one legitimately named `Unknown-Something` alone.
  Both surfaces are checked.
- **`third-party-notices` is a prerequisite of `check-third-party-notices`**, not a `$(MAKE)` call from
  inside the recipe. Regeneration therefore runs before the tracked-file check
  rather than after it, so an untracked file costs a regeneration before
  failing. The tracked check still precedes the `git diff`, because `git diff`
  reports nothing for an untracked path.

## Scope

Runtime is the import set of `./cmd/...`, a wildcard over `./cmd/*`:
`make PREFIX=/artifacts cmds` builds both binaries, the image copies them in, and
the systemd packages ship the same two. Against `./...` the set is identical (396
non-stdlib packages on both platforms), so nothing is dropped, and it still
tracks the shipped set if `./examples` later grows a dep.

The second surface is `nvidia-ctk`, pinned in `deployments/devel/tools.go` and
copied into the released image as `/usr/bin/nvidia-ctk`, so its dependencies are
redistributed. `moq` and `go-licenses` are pinned in the same file but run only
at build time and ship nowhere, so they are excluded. That drops 67 packages
that no released artifact contains; 9 more of their dependencies are shared with
the runtime surface and remain attributed there. A tool added to `tools.go`
later is included by default, since exclusion is an explicit list, so
over-attributing is the failure mode rather than a silent omission.

Platforms are `linux/amd64` and `linux/arm64`, checked against `multi-arch.mk`.
The bundled-binary pass uses the same two platforms.

Not covered: the non-Go contents of the released image, the distroless base and
the BusyBox tree copied in, which have their own compliance processes. The
document says it covers Go deps only.

## Shipping

Committed to the repository. Not copied into the image or the packages, so their
contents are unchanged, and not attached to releases, so the release flow is
untouched.

The check has no changed-paths filter. The list is an import set, so it
goes stale when a `.go` file changes its imports, not only when `go.mod` or
`vendor/` move.

## Testing

- **Cross-host.** Same sha256
  `def935e1ce73562433fccfc901342134ca0dde616aa2b8992774c8111f341c81` on
  macOS/arm64 (BSD awk, BSD sort) and in `golang:1.26.5` (mawk 1.3.4, GNU sort),
  where mawk showed the array-instantiation difference.
- **Determinism.** Two runs on one host are byte-identical.
- **Completeness.** Expected set from `go list -deps ./cmd/...` on both
  platforms, stdlib and local module dropped, mapped to license dirs by longest
  prefix: 373 runtime packages plus 113 behind `nvidia-ctk`, 474 unique, 0
  missing and 0 unused rows. The `moq` and `go-licenses` binaries are reachable
  from neither shipped surface.
- **Content.** All 77 index rows have a license-text section. 0 `Unknown`, 0
  "License text unavailable", 0 unresolved `module@version`.
  `filepath-securejoin` shows `BSD-3-Clause / MPL-2.0` with `COPYING.md`,
  `LICENSE.BSD` and `LICENSE.MPL-2.0`.
- **`make check-third-party-notices`** passes on the committed file, and fails both when it
  is stale and when it is untracked.
- **Error paths.** Missing input file, unreadable or empty `vendor/modules.txt`,
  and a platform matrix out of sync with `multi-arch.mk` (added `linux/ppc64le`;
  the run failed with both lists printed).
- **Lint.** `shellcheck` clean. On the modified workflow `actionlint` reports
  only the `linux-amd64-cpu4` unknown-runner-label warning and the SC2086
  pattern, both copied from the sibling jobs and both pre-existing elsewhere in
  the file. `yamllint -d relaxed` reports
  only line-length warnings.
- **Not verified.** The workflows have not run on NVIDIA's runners, so
  the check reaching the module hosts through the Artifactory Go proxy is
  untested. The generator resolves the bundled binary's source URLs over the network
  (`?go-get=1`) and fails loudly rather than emitting `Unknown`, so a blocked
  host is a failed job, not a degraded file.


.github/workflows/notices-check.yaml
```diff
@@ -0,0 +1,60 @@
...
+# committed file goes stale, and the failure would surface on main instead of on
+# the pull request that caused it.
+
+name: notices
+
+on:
+  push:
+    branches:
+      - main
+      - "pull-request/[0-9]+"
```
rajathagasthya: Why does this need to run for a pull request?

.github/workflows/notices-check.yaml
```diff
@@ -0,0 +1,60 @@
...
+# copy, so a dependency change cannot land without refreshed attribution.
+#
+# This runs on every build rather than behind a changed-paths filter. The
+# runtime inventory is the import closure of ./cmd/..., so it changes when
+# ordinary .go files change their imports, not only when go.mod or vendor/ move.
+# A filter keyed on dependency manifests would therefore pass green while the
+# committed file goes stale, and the failure would surface on main instead of on
+# the pull request that caused it.
+
+name: notices
```
rajathagasthya:
```suggestion
name: third-party-notices
```
Also rename the file.

hack/generate-notices.sh
```diff
@@ -0,0 +1,526 @@
+#!/usr/bin/env bash
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
rajathagasthya:
```suggestion
# Copyright (c) NVIDIA CORPORATION.  All rights reserved.
```
Let's remove the year from copyright everywhere.

.github/workflows/release-third-party-notices.yaml
tariq1890, line 1: As discussed offline, let's not publish the TPN as its own release asset.

.github/workflows/notices-check.yaml
tariq1890: Shouldn't this be run as a sub-workflow instead? That way, it'll just inherit the same `on:` conditions as the main CI check.

That's what is being followed in the k8s-device-plugin IIUC - https://github.com/NVIDIA/k8s-device-plugin/pull/1950/changes

Makefile
```diff
@@ -81,6 +81,35 @@ vendor:
...
+# the modules linked into the released binaries and the build toolchain.
+notices: $(GO_LICENSES)
+	@bash hack/generate-notices.sh
+
+# Verify THIRD_PARTY_NOTICES.md is in sync with the dependency tree.
+notices-check:
+	@echo "- Checking if THIRD_PARTY_NOTICES.md is up to date..."
+	@# Cheap gate first, before spending minutes regenerating. git diff reports
+	@# nothing for an untracked path, so an uncommitted notices file would
+	@# otherwise pass this gate silently.
```
tariq1890: Please make sure that these auto-generated comments are removed. Most if not all comments generated in this PR are likely not needed

.github/workflows/notices-check.yaml
abrarshivani: @rajathagasthya We need to run this for every pull request because a PR can introduce or update a third-party library, which may change the contents of the third-party notices file. Running this check as part of every PR ensures that the notices file stays up to date with all third-party dependency changes.
@tariq1890 Make sense, will update this.

.github/workflows/notices-check.yaml
abrarshivani: Will update this

hack/generate-notices.sh
abrarshivani: Before updating, just want to confirm whether this comply with NVIDIA's copyright-header policy?

.github/workflows/release-third-party-notices.yaml
abrarshivani: sure, will update this

.github/workflows/notices-check.yaml
tariq1890: > Running this check as part of every PR ensures that the notices file stays up to date with all third-party dependency changes.

I agree with this. My point was that we should just add this check as a part of the existing CI pipeline which runs for every PR and every push to `main`. I don't see a need to define a new workflow with its own `on:` conditions, we can just use the existing workflow, yes?

Makefile
abrarshivani: Sure, will remove the comments which are not needed.

hack/generate-notices.sh
tariq1890: Yes, the copyright header is compliant and the one without the year is actually recommended template

abrarshivani: Thanks for confirming. Will update this.

.github/workflows/notices-check.yaml
abrarshivani: Yes, we can use existing workflow, I plan to update the same.

abrarshivani:
@rajathagasthya @tariq1890 I have addressed the comments. Can you please take another look?

.github/workflows/golang.yaml
```diff
@@ -86,3 +86,30 @@ jobs:
...
+      uses: actions/setup-go@v7
+      with:
+        go-version: ${{ env.GOLANG_VERSION }}
+    - name: Setup Go Proxy
+      id: setup-go-proxy
+      uses: nv-gha-runners/setup-artifactory-go-proxy@main
+    - env:
+        GOPROXY: ${{ steps.setup-go-proxy.outputs.goproxy-url }}
+      run: |
+        make notices-check
```
tariq1890:
```suggestion
        make check-third-party-notices
```

hack/generate-third-party-notices.sh
tariq1890, line 1: rename to `generate-third-party-notices.sh` please

.gitignore
```diff
@@ -2,3 +2,7 @@
 *.swo
 dist/
 .vscode/
+# Build tools installed by 'make notices'
```
tariq1890:
```suggestion
```

.gitignore
```diff
@@ -2,3 +2,7 @@
...
+/bin/
+# Scratch tree of collected dependency licenses (see hack/generate-notices.sh)
```
tariq1890:
```suggestion
```

Makefile
```diff
@@ -81,6 +81,23 @@ vendor:
 check-vendor: vendor
 	git diff --quiet HEAD -- go.mod go.sum vendor
 
+GO_LICENSES = $(CURDIR)/bin/go-licenses
+
+$(GO_LICENSES): deployments/devel/go.mod deployments/devel/go.sum
+	cd deployments/devel && GOBIN=$(CURDIR)/bin GOFLAGS=-mod=readonly go install github.com/google/go-licenses/v2
+
+notices: $(GO_LICENSES)
```
tariq1890:
```suggestion
third-party-notices: $(GO_LICENSES)
```

Makefile
```diff
@@ -81,6 +81,23 @@ vendor:
...
+	@bash hack/generate-notices.sh
+
+# git diff reports nothing for an untracked path, so the tracked check has to
+# run before it or an uncommitted notices file would pass silently.
+notices-check: notices
```
tariq1890:
```suggestion
check-third-party-notices: third-party-notices
```

Makefile
```diff
@@ -81,6 +81,23 @@ vendor:
...
+	@echo "- Checking if THIRD_PARTY_NOTICES.md is up to date..."
+	@git ls-files --error-unmatch THIRD_PARTY_NOTICES.md >/dev/null 2>&1 \
+		|| { echo "ERROR: THIRD_PARTY_NOTICES.md is not tracked. Run 'make notices' and commit the result."; exit 1; }
```
tariq1890:
```suggestion
		|| { echo "ERROR: THIRD_PARTY_NOTICES.md is not tracked. Run 'make third-party-notices' and commit the result."; exit 1; }
```

Makefile
```diff
@@ -81,6 +81,23 @@ vendor:
...
+	@git diff --exit-code -- THIRD_PARTY_NOTICES.md \
+		|| { echo "ERROR: THIRD_PARTY_NOTICES.md is stale. Run 'make notices' and commit the change."; exit 1; }
```
tariq1890:
```suggestion
		|| { echo "ERROR: THIRD_PARTY_NOTICES.md is stale. Run 'make third-party-notices' and commit the change."; exit 1; }
```

tariq1890 (dismissed):

hack/generate-third-party-notices.sh
```diff
@@ -0,0 +1,473 @@
...
+    ( cd "${TOOLS_DIR}" && GOFLAGS="-mod=readonly" go mod download ) >&2
+
+    for platform in "${PLATFORMS[@]}"; do
+        goos="${platform%/*}"
+        goarch="${platform#*/}"
+        log "Collecting bundled binary licenses for ${goos}/${goarch}..."
+
+        save_dir="${SAVE_ROOT}/tools/${goos}_${goarch}"
+        (
+            cd "${TOOLS_DIR}"
```
rajathagasthya, line 166: Maybe I'm missing something, but why do we need to run this in the tools directory?

abrarshivani: We run it from tools directory because `nvidia-ctk` is pinned in that Go module, not the root module, so `go-licenses` needs that module context to resolve its dependencies. We include `nvidia-ctk` because it ships in the image, while `moq` and `go-licenses` are excluded since they're build time only tools.

rajathagasthya (approved):

abrarshivani:
@tariq1890 Can you take another look? I have updated the generation script to include only modules that are redistributed in the release image from devel/go.mod, which is nvidia-ctk, while moq and go-licenses are excluded since they're build time only tools.

hack/generate-third-party-notices.sh
```diff
@@ -0,0 +1,473 @@
...
+* **Bundled binaries** — Go modules behind `nvidia-ctk`, which is built from the
+  version pinned in `deployments/devel/go.mod` and copied into the released image
+  as `/usr/bin/nvidia-ctk`. `moq` and `go-licenses` are also pinned there but run
+  only at build time and are not redistributed, so they are not listed.
+
+Go standard library packages are excluded; they are covered by the license of
+the Go distribution itself.
+
+Scope: this document does not inventory the non-Go contents of the released
+container image — the packages inherited from its distroless base image or the
```
tariq1890: All of these are implementation details that don't need to be mentioned in the doc. We should follow the same approach as what was done in the nvidia-container-toolkit repo. So the intro should be similar to this

```
NVIDIA Container Toolkit

This file lists every third-party dependency that NVIDIA Container Toolkit redistributes, along with the verbatim text of each dependency's license. In particular, this covers all Go modules statically linked into the commands under cmd/. The nvidia-container-runtime-hook, nvidia-container-runtime, nvidia-container-runtime.cdi, nvidia-container-runtime.legacy, nvidia-ctk and nvidia-cdi-hook commands ship in the deb and rpm packages. The nvidia-ctk-installer command ships in the container-toolkit image. Go standard library packages are excluded; they are covered by the license of the Go distribution itself.

The container-toolkit image uses nvcr.io/nvidia/distroless/go as a base image. All of the OSS packages and source included in this image can be found at https://developer.nvidia.com/w/distroless-oss/index.html. A statically compiled busybox binary is added to the image, which is licensed under GPLv2.
```

hack/generate-third-party-notices.sh
```diff
@@ -0,0 +1,473 @@
...
+BusyBox tree copied in for shell utilities. Those are covered by their own
+compliance processes, including any source-distribution obligations they carry.
+
+Go runtime entries identify their dependency as `module@version` rather than a
+URL. `go-licenses` in vendor mode reports a link into this repository at `HEAD`,
+which stops describing the released content once `main` advances;
+`module@version` is immutable and names the upstream project directly.
```
tariq1890: These are implementation details too; they shouldn't show up in the final markdown doc either

THIRD_PARTY_NOTICES.md
```diff
@@ -0,0 +1,11004 @@
...
+All of the OSS packages and source included in this image can be found at
+<https://developer.nvidia.com/w/distroless-oss/index.html>. A statically
+compiled busybox binary is added to the image, which is licensed under GPLv2.
+
+Entries name their dependency as `module@version` rather than a URL, because
+`go-licenses` in vendor mode links into this repository at `HEAD`, which stops
+describing released content once `main` advances.
+
+This file is generated. Run `make third-party-notices` to regenerate it;
+`make check-third-party-notices` verifies it is current.
```
tariq1890: This needs to be removed as well

rajathagasthya:
@abrarshivani Once this is merged, you will need to manually backport this to `release-0.14` branch. Auto backport with `/cherry-pick` doesn't work when workflow files are changed (see https://github.com/NVIDIA/mig-parted/pull/457#issuecomment-5269865687).

abrarshivani:
@rajathagasthya Sounds good, will backport once merged.

tariq1890 (approved):

PR #467, opened 2026-08-14
Manage the Go toolchain version with Renovate

rajathagasthya:
Dependabot's docker manager treats Go's RC docker tags (e.g. `1.27rc2`) as the latest version and stops proposing stable patch updates once the RC update is declined.

Run a self-hosted Renovate daily with regex managers backed by the `golang-version` datasource, which only serves stable Go releases. One grouped PR updates both the golang bootstrap image in the devel Dockerfile and the `GOLANG_VERSION` build-arg default in the container Dockerfile; the latter was stale at 1.22.8 because real builds always override it. Drop the dependabot docker entry for `deployments/devel` so the two bots do not race on the same line.

Renovate manages Go version in GPU Operator, so this matches it.

### DIFF
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/checkout@v7
 
       - name: Run Renovate
-        uses: renovatebot/github-action@v46.1.21
+        uses: renovatebot/github-action@v46.2.2
         with:
           configurationFile: .github/renovate.json
         env:
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -1,0 +1,22 @@
+name: Renovate
+
+on:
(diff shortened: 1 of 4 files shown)


.github/workflows/renovate.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  renovate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Run Renovate
+        uses: renovatebot/github-action@v46.1.21
```
rahulait: nit: lets use v46.2.2 version

rahulait (approved):
Just one nit, otherwise LGTM

.github/workflows/renovate.yaml
rajathagasthya: Done.

PR #473, opened 2026-08-19
[deps] bump x/net, x/text and x/mod to newer versions

rahulait:
Bump to newer versions to fix CVEs


deployments/devel/go.mod
```diff
@@ -39,12 +39,12 @@ require (
 	github.com/urfave/cli-altsrc/v3 v3.1.0 // indirect
 	github.com/urfave/cli/v3 v3.10.1 // indirect
 	go.opencensus.io v0.24.0 // indirect
-	golang.org/x/mod v0.38.0 // indirect
-	golang.org/x/net v0.56.0 // indirect
-	golang.org/x/sync v0.21.0 // indirect
+	golang.org/x/mod v0.40.0 // indirect
```
rajathagasthya, line 42: Are these indirect dependencies of toolkit and show up as having CVEs? Wondering why we need to bump deps in `devel/go.mod` urgently.

rajathagasthya (approved):

deployments/devel/go.mod
rahulait: container-toolkit uses x/mod v0.38.0 and I am trying to bump it just so that any binary built and shipped should use updated package without CVEs.

rahulait: It doesn't fixes any CVE though as no binaries are built and shipped with this, so we can wait to merge this later on as well.

PR #489, opened 2026-09-05
fix: ignore zero-count profiles when comparing MIG configs

git-jxj:
MIG configs accept zero-count profiles and skip them when creating
instances. The resulting device configuration omits those profiles, but
`Equals` compares map lengths and treats the configurations as different.

For example, applying `{"1g.5gb": 2, "2g.10gb": 0}` produces
`{"1g.5gb": 2}`. This causes `assert` to report a mismatch and prevents
`apply` from skipping an already configured GPU.

Compare profile counts in both directions, treating omitted profiles as
zero. Add equality tests and a set/get regression using the existing
DGX A100 NVML mock. Three equality cases and the set/get regression fail
against unmodified main and pass with the fix.

Validation:
- `go test -race ./... -count=1`
- `make test`
- `make lint` (0 issues)
- CLI `assert --valid-config` with a zero-count profile

The CLI validation uses a real config file. Applying and reading the
device configuration use the project's NVML mock; no physical MIG-capable
GPU was used.


rajathagasthya (approved):
Thanks @git-jxj! This looks like a real improvement to avoid unnecessary mismatch and recreating instances. 

Let's get another approval before merging.

rahulait (approved):
LGTM

